PR #21, opened 2015-01-15
yaboot needle

k0da:
Signed-off-by: Dinar Valeev <dvaleev@suse.com>


coolo:
the needle is sle11 specific and doesn't belong here

PR #22, opened 2015-02-17
Detect remapped shortcuts for other desktops

DimStar77:


DimStar77:
This was needled on openQA.o.o

PR #23, opened 2015-03-12
Needles for SSSD test suite

HouzuoGuo:
reference
https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/317


HouzuoGuo:
Will use alternative "validate_script_output" instead of screen needles.

PR #24, opened 2015-03-19
inst-consolesettingstyped for uefi

lnussel:


coolo:
do we already exclude ARCH needles?

lnussel:
Stephan Kulow schrieb:
> do we already exclude ARCH needles?

Nope. I added the tag just in case.

PR #27, opened 2015-05-20
Needle for kde-ready desktop (used in second_boot test case)

HouzuoGuo:


coolo:
we should avoid the tag 'kde-ready'. But rather check for desktop-at-second-boot or even better generic-desktop and have the needle then have the ENV-DESKTOP=kde set.

PR #38, opened 2015-07-07
[do not merge] Add needles for yast2-users on ncurses

imobach:
Needles for [yast2-users tests PR](https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/533).


coolo:
I don't mind merging needles early, but I would prefer if your needle filenames contained the timestamp they were created. The webui will replace that timestamp if we have to in the future

imobach:
Ok, no problem. I've renamed all the needles adding the timestamp. I'll take that into account in the future.

Thanks!

PR #42, opened 2015-08-13
Needles for nfs-client test and alternative for dns-server (change in font)

ancorgs:


sysrich:
Please also consider submitting SLE12 needles :)

PR #50, opened 2015-12-18
jeos needles

lnussel:


coolo:
then we need the same unregister call in main.pm as on sle12 branch

lnussel:
yes, that's https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/902

PR #52, opened 2015-12-30
renamed inst-live-* to boot-live-* to avoid confusion

nilxam:


nilxam:
required by https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/909

PR #54, opened 2016-01-20
Correct check console font by narrowing

okurz:
This way the needle can actually match on the echoed line, not the input line.
Related to https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/947
but not actually a requirement.

Crosschecked with rbrown.


okurz:
@lnussel you might be interested in this one and https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/947

PR #53, opened 2016-01-12
Add needles for awesome test

okurz:
Related to issue: https://progress.opensuse.org/issues/9522

PR #55, opened 2016-01-22
Add needles for awesome test

asdil12:
Related to issue: https://progress.opensuse.org/issues/9522

Obsoletes  #53 

Required by os-autoinst/os-autoinst-distri-opensuse#940


coolo:
No need to obsolete PRs, they follow your branch

coolo:
what's missing now is unregistering the needles these needles for minimalx. Actually it would be an idea to leave minimalx needles for DESKTOP=awesome.

coolo:
this way you avoid needling everything twice

okurz:
PR #53 was mine so I think @asdil12 has to create a new one unless we use a common feature branch on the main repo itself which we overwrite.
Regarding the unregistering there might be a problem with the DESKTOP variable being reused ?!?

coolo:
I didn't say it's easy - but it's worthwhile

asdil12:
Reusing minimalx needles for
- inst_overview
- displaymanager

PR #61, opened 2016-02-24
Needles for iscsi server/client test suite

dzedro:
Tumbleweed:
server: http://10.100.98.90/tests/1174
client: http://10.100.98.90/tests/1175


okurz:
what do you need the gnome-runner and clean-desktop needles for?

dzedro:
before i started a pulled all needles and this two needles were missing I had to create them in gnome
they are part of x11_start_program

dzedro:
I used DESKTOP=gnome when I created openSUSE image, no idea what DM is default for openSUSE, maybe KDE.

okurz:
ok, I see. @DimStar77 @sysrich @nilxam could you take a look, see discussion, please?

DimStar77:
imho, desktop-runner-gnome and generic desktop should not be part of any test-specific submission. They can 'change over time', depending on the versions tested (it's not uncommon for the UI styles to change). Every single test running X11 apps makes use of those two needles anyway, so whenever they need a fix, we are sure to introduce them.

nilxam:
your needles seems outdated, currently the matched needle on openqa.oo is consoletest_finish-gnome-TW-20151222 https://openqa.opensuse.org/tests/123693/modules/xterm/steps/4/edit

and *even more*, I guess you didn't set DESKTOP variable in your job settings, then DESKTOP will be kde as default fallback, so all GNOME needles will not load due to cleanup_needles().

dzedro:
I used DESKTOP=gnome to build that disk image, and I have this needle. http://10.200.98.90/tests/1078/modules/first_boot/steps/2
I didn't define DESKTOP for client test http://10.100.98.90/tests/1175

I will remove this two needles

nilxam:
yeah, your server/client test need to define DESKTOP=gnome as well, http://10.100.98.90/tests/1175/modules/boot_to_desktop/steps/2/edit proves it is loaded KDE needles :)

PR #66, opened 2016-03-23
Add workaround needle for multi_users_dm-xfce

asdil12:
This reverts commit 23db8e1635886f205eff0e3ac625e416d32b7243.

There is no needle for the xfce login manager by intention
because the usability for the login manager is bad when
there are many users in passwd file.

See https://progress.opensuse.org/issues/9694
and https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/1102

The test is meant to fail for xfce until the login manager is fixed.
This shouldn't be a problem as it isn't marked as important.


coolo:
where is your bug report about it?

coolo:
If you have one, rename the needle to include that bug number and add a workaround property. A missing needle really doesn't document much

asdil12:
Why would I need a bug report for that?

asdil12:
I don't see the need of a workaround needle.
This test is not flagged as important so it won't affect the overall result.
So it can fail without affecting other tests.

coolo:
yes, but it affects every day's review.

coolo:
and if you don't feel the need to create a bug report - why did you feel the need to write a test that fails?

sysrich:
Ping - are you going to file a bug about this?

asdil12:
https://bugzilla.opensuse.org/show_bug.cgi?id=981596

okurz:
there is a bug and a test, can you make a workaround needle according to https://progress.opensuse.org/projects/openqatests/wiki#applying-quotworkaroundquot-needles ?

asdil12:
Converted the existing needle into a workaround needle

sysrich:
Can you edit the PR subject to match the new purpose of this PR - just so no one gets confused after we merge it ;)

asdil12:
done

PR #75, opened 2016-05-31
for overwrite

coolo:


okurz:
this was some test regarding picture comparison, right?

PR #80, opened 2016-06-20
Add screenlock and its password needles for opensuse13.2

mitiao:
From the log of this failure, it looks like there were no needles of tag 'screenlock' and 'gnome-screenlock-password' for opensuse13.2:
https://openqa.opensuse.org/tests/212630#step/consoletest_finish/8

And verified result here:
http://147.2.207.208/tests/1402/modules/consoletest_finish/steps/5


gnome-screenlock-password-opensuse132-20160620.json
```diff
@@ -0,0 +1,15 @@
...
+    {
+      "type": "match",
+      "ypos": 237,
+      "height": 263,
+      "width": 458,
+      "xpos": 290
+    }
+  ],
+  "tags": [
+    "gnome-screenlock-password"
```
okurz: shouldn't these tags have the additional tags as screenlock-gnome-opensuse132-20160619.json, too?
@sysrich, @coolo

screenlock-gnome-opensuse132-20160619.json
```diff
@@ -0,0 +1,25 @@
...
+      "height": 63,
+      "type": "match",
+      "ypos": 320
+    }
+  ],
+  "tags": [
+    "screenlock",
+    "ENV-DESKTOP-gnome",
+    "ENV-DISTRI-opensuse",
+    "ENV-INSTLANG-en_US"
```
okurz: shouldn't these tags have ENV-VERSION-13.2, too?

mitiao: added. this tag has not been treated in opensuse/main.pm, maybe for future?

gnome-screenlock-password-opensuse132-20160620.json
okurz: mitiao, please add the corresponding tags as in screenlock-gnome-opensuse132-20160619.json, i.e ENV-DESKTOP-gnome, ENV-DISTRI-opensuse, ENV-VERSION-13.2 as the needle already has the specific requirements in the name.

coolo: I don't think you need DESKTOP-gnome in a needle that already has gnome in its tag. And the DISTRI-opensuse part is really obsolete as we only have one needle repo for it. But the VERSION-13.2 one will make it possible not to check it on any other version (which I think is not yet implemented in main.pm)

okurz: err, what do mean "it is not implemented"? Isn't ENV-VERSION-13.2 automagically set based on VERSION in `vars.json`? somewhere? anywhere? magic?

coolo: https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/products/sle/main.pm#L107 - something like that does not exist in opensuse/main.pm (but it should)

okurz: Why this blacklisting and not whitelisting? Can this be done automatically for all ENV variables?

lnussel:
The png seems to be corrupted somehow:
Could not open image /var/lib/openqa/share/tests/opensuse/products/opensuse/needles/gnome-screenlock-password-opensuse132-20160620.png
DIE tinycv::Image::search_needle: needle is not of type tinycv::Image at /usr/lib/perl5/vendor_perl/5.18.2/x86_64-linux-thread-multi/tinycv.pm line 87.

lnussel:
it's json

okurz:
On Monday 20 June 2016 13:17:50 Ludwig Nussel wrote:
> it's json

so we learned: if github can not display the image, something might be wrong 
with the image, not github …

PR #84, opened 2016-07-20
Add inst-bootmenu tag to existing Leap 42.2 needle

okurz:


coolo:
oh no. even more bootmenus that are checked for all opensuses alike? ;(

PR #89, opened 2016-07-24
Delete firefox-reader-window replaced by more specific

okurz:
Reader window uses 'assert_and_click' which never works unless the 'x' is
clicked, which it should be in the new needle.


nilxam:
fine with me. but please note that, assert_and_click "firefox-reader-view" is not mainly to close the reader view window, but just click on it. the reason: when firefox just introduced reader view feature in previous firefox version, it has a strange behavior - the reader view window grabbed focus than the firefox main window, so the main window's keyshortcut was ignored ie. alt-h will not show the help list, have to click on reader view window or close it than the focus will back to main window.

PR #92, opened 2016-07-31
Add needles for Krypton as of 2016-07-31

okurz:


generic-desktop-kde-live-20160731.json
```diff
@@ -0,0 +1,15 @@
+{
+  "tags": [
+    "generic-desktop"
```
coolo: Shouldn't this be limited to FLAVOR-Krypton as well?

okurz: I don't think so. I missed ENV-DESKTOP-kde however. Limiting to Krypton would be counter productive as we can probably expect any plasma related needle in Krypton to hit TW about 2-6 weeks later or not?

okurz:
added boot menu needle

generic-desktop-kde-live-20160731.json
coolo: but not Leap - and you match TW branding also on leap. we already have quite some generic-desktop needles ;(

okurz: hm, ok, so ENV-LEAP-0. It should still match FLAVOR-Argon, though, right?

okurz: maybe you can help with https://progress.opensuse.org/issues/12956#open-questions

sysrich:
the generic-desktop needle seems suspicious - I'd expect a KDE live desktop to have a Desktop Folder, like https://openqa.opensuse.org/tests/232774#step/first_boot/1 - even on Krypton

okurz:
removed the desktop needle without icons

PR #93, opened 2016-08-05
 New needles for Activate existing encrypted volume

richiejp:
Also includes some new needles for existing tests which did not have a
current needle or have never had one for OpenSUSE.


first_boot-displaymanager-20160805.json
```diff
@@ -0,0 +1,15 @@
+{
+  "properties": [],
+  "tags": [
+    "displaymanager"
```
coolo: why do you need a new displaymanger needle?

user_settings_root-rootpassword-typed-20160805.json
```diff
@@ -0,0 +1,15 @@
+{
+  "tags": [
+    "rootpassword-typed"
```
coolo: to me this looks like you recreated all needles

first_boot-displaymanager-20160805.json
richiejp: OpenQA failed to match on this needle, I searched the needle directory and could not find an existing one.

user_settings_root-rootpassword-typed-20160805.json
richiejp: I have checked and none of the needles previously existed.

first_boot-displaymanager-20160805.json
okurz: the tags don't relate to filenames. If searching for existing ones in the workspace of needles, better use `git grep <tag>`, e.g. `git grep displaymanager` which should show you the needle filenames mentioning this tag, e.g.

```
$ git grep displaymanager
change_password-displaymanager-20160520.json:    "displaymanager"
desktop-wait-to-login-after-virt-test-20160519.json:    "virttest-displaymanager"
displaymanager-20140621.json:    "displaymanager",
displaymanager-20160502.json:    "displaymanager",
displaymanager-20160506.json:    "displaymanager",
displaymanager-password-prompt-20160517.json:    "displaymanager-password-prompt"
displaymanager-password-prompt-gnome-SP1-20160519.json:    "displaymanager-password-prompt"
displaymanager-settings-20160322.json:    "displaymanager-settings"
displaymanager-settings-20160520.json:    "displaymanager-settings"
displaymanager-sles4sap-nousers-20160513.json:    "displaymanager",
first_boot-displaymanager-password-prompt-minimal-20160517.json:    "displaymanager-password-prompt"
gdm-workaround-bsc962806-20160125.json:    "displaymanager",
xdm-20150429.json:      "displaymanager",
xdm-20160504.json:    "displaymanager",
```

okurz: We might be able to help you if you point us to your local web instance where the needle you created is matched and no other is found.

richiejp: I did that and checked the needles with that tag and they all appeared to be wrong. However I have checked again and found that my working directory is a bit out of date and that someone has since added new gdm needles which should match for my scenario: https://github.com/os-autoinst/os-autoinst-needles-opensuse/commit/18fa6eb19e7acca2cabebaaaa21ee568e91e3fb0

richiejp: Indeed the gdm needle does work with my test. I simply can't find any existing needle tagged with "rootpassword*" though, which I guess is the only other one you find suspicious?

okurz: We find any needle suspicious that we expect to be not related to what you added. If you didn't change anything about gdm then there shouldn't be a needle in this PR for gdm. If the root password was already typed for all the tests in before, then there shouldn't be a needle in this PR. *If*, however, the tests *always* worked with creating a new user *and* having the checkbox "use same password for system administrator" checked, *then* we might actually never needed the needles and need them now. If you can confirm this is the case then of course, keep it.

okurz: I just checked test_suites and job_groups on o3 and found no Tumbleweed scenario which is setting DOCRUN so the root password is never actually typed.

@richiejp I guess you set DOCRUN to make it work because otherwise user_import does not work because it does not expect the root password screen? If this is the case, please adapt the check in `products/opensuse/main.pm` to check for your new variable, too.

*and* we should think about why the test_suites `nice` and `doc` are not run anywhere.

richiejp: Yes and NOAUTOLOGIN. I will check for my variables instead then.

PR #94, opened 2016-08-12
Fix packagekit updates

kravciak:


generic-desktop-kde-after-update-20160810.json
```diff
@@ -0,0 +1,25 @@
...
+      "xpos": 0,
+      "width": 61,
+      "height": 26
+    }
+  ],
+  "tags": [
+    "ENV-DESKTOP-kde",
+    "ENV-DISTRI-opensuse",
+    "ENV-VERSION-42.1",
+    "generic-desktop"
```
coolo: This whole KDE on 42.1 is a mess ;(

PR #97, opened 2016-08-15
Delete wrong "updates available needles"

okurz:
* "shield icon with green checkmark" --> *no* updates available
* "shield icon with red 'x'" --> updates *are* available


okurz:
Looks like the icon can also be green although updates are still available: http://imagebin.suse.de/2093/img
mkravec confirmed that the green icon does not go away just by waiting. "green" could also mean "only security fixes".

okurz:
if we have security type update, the icon will be security-low, if important type then security-medium, otherwise security-high so both icon are means have update available

kravciak:
I could be wrong, but from my observation:
no icon = no updates
green = no security updates
red = security updates

nilxam:
basically right,

QString PkUpdates::iconName() const
{
    if (securityCount() > 0)
        return "security-low";
    else if (importantCount() > 0)
        return "security-medium";

    return "security-high";
}

the icon with red x and green v both means have updates available.
no icon in the system tray means the system is updated, but the icon still be with green v, just if the systemtray's function is worked properly(since updater is auto show/hide) it does not show up in the systemtray.

so we need both green v and red x needle.

PR #98, opened 2016-08-15
Correct OWF->OFW and delete again wrong needle

okurz:


okurz:
@k0da please don't cover needles in https://progress.opensuse.org/projects/openqatests/wiki/Wiki#don39t-include-version-specific-content-in-needles
Also, you recreated a needle that I already did but both of us didn't notice because there was a typo in the name 'OFW'

PR #102, opened 2016-08-17
Fix typo - correction to commit 68eb12 (file renamed, tag uncorrected)

DimStar77:


okurz:
interesting I missed this

PR #109, opened 2016-09-29
no workaround in desktop-selection-text-eurlatgr-20160918.json

michelmno:
to avoid to report test as "soft failed" in
https://openqa.opensuse.org/tests/272409#step/logpackages/1


okurz:
seems, correct. If I followed correctly, @lnussel aka. "our wrong font investigator", accepted the font and renamed the needle but forgot to remove the workaround property.

okurz:
BTW why do you always sign off your own commits?

PR #111, opened 2016-10-04
xterm and sshxterm needles updates for PowerPC

michelmno:


okurz:
also here, why should powerPC need extra needles? The default ones should apply regardless of the architecture, isn't it?

michelmno:
I created a new needle for xterm because the xterm-20140710 only match 95% and is reported as red.
I reduced the length of the matching area to not compare the empty space after the dot,
and this is sufficient to match 100%
May be I could modify the json of the original needle, but did not wanted to impact previous tests.
```
   "area": [
     {
       "height": 14,
       "xpos": 16,
-      "ypos": 207,
+     "ypos": 214,
       "type": "match",
-      "width": 276
+     "width": 254,
     }
   ]
```

Similarly, I created a new  sshxterm needle by comparing the two lines by two different areas, without the trailing empty space after dots; rather to have only one comparison for the two lines with empty space after dots.

okurz:
I am sorry but I don't yet understand what you want to achieve. Is this about fixing test failures in [openSUSE Tumbleweed PowerPC](https://openqa.opensuse.org/group_overview/4) ? If yes, please reference the according jobs that fail and would be fixed but currently I don't see any failures in https://openqa.opensuse.org/tests/overview?distri=opensuse&version=Tumbleweed&build=20160927&groupid=4 but I also don't see any tests that run xterm and sshxterm tests. I wonder if we should enable a job for this if this is what you intend to fix. AFAIK @k0da was interested in keeping the openSUSE PowerPC tests on openqa.opensuse.org green. You are welcome to also join the irc channel irc://chat.freenode.net/opensuse-factory for discussion

michelmno:
I am locally trying tests for PowerPC that are not covered yet for PowerPC in your above url.

okurz:
That makes sense. Thank you very much already. We can merge the needles but keep in mind that we clean up needles, I.e. delete them, when they aren't used for some time on the productive instance.

PR #112, opened 2016-10-04
accept-ssh-host-key-20161004 for PowerPC

michelmno:


okurz:
Can you explain why PowerPC needs to handle keys differently than other tests?

michelmno:
I created a new needle only because the delimited area that is capturing the last line
is partially grabing the bottom of previous line containing the key fingerprint that differ in my case.
So I am reducing the height in the new needle:
```
   "area": [
     {
       "width": 434,
-      "ypos": 287,
-      "height": 19,
+      "ypos": 289
+      "height": 17,
        "xpos": 0,
       "type": "match",
```

okurz:
I see, makes sense. Thank you.

PR #114, opened 2016-10-06
yast2_snapper-snapshots-20161006 new needle

michelmno:
this new needle for default kde desktop,
Yast2 Snapper window is positionned differently compared to gnome
so compared areas not the same versus yast2_snapper-snapshots-gnome*

Signed-off-by: Michel Normand normand@linux.vnet.ibm.com


okurz:
The window position does not matter the way the matching works. The position is basically an "initial" position but the match is tried on all possible locations on the screen. What matters though is the font which very well might differ between gnome and kde.

The problem I see with this proposed needle change is that I don't see the snapper test run anywhere on openqa.opensuse.org so I am not sure what to think about this PR and where we would need it.

michelmno:
closing this PR for the moment, because related test ran on an openQA server that is not openqa.opensuse.org , and do not have access on this last one to execute related "client jobs" command.

PR #116, opened 2016-10-08
update needle for gdm_session_switch

chuchingkai:
Failed at: https://openqa.opensuse.org/tests/275687#step/gdm_session_switch/29

I updated the needle for generic desktop screenshot for sle-classic


coolo:
you remove the exclude area. Are you aware? exclude areas aren't really well covered in the webui needle editor

chuchingkai:
@coolo Yes, that's intentional. The new needle is more accurate so it doesn't need exclude area any more.

PR #119, opened 2016-10-10
Remove useless 'no updates detected' needles

DimStar77:
It is not reliably detectable that there is no icon on the tray (not
detecting the icon could just mean the icon changed)

It was decided to no longer test for updates in upgrade tests


DimStar77:
Goes hand-in-hand with https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/1940
but even without that: the detection is unreliable as the icons are not predictably sorted on the tray... and with so many icons in upgrade scnearios it's not realistic to match all options

PR #121, opened 2016-10-25
Add needle for 'yast2 finished'

okurz:


okurz:
disappointed, very disappointed

PR #127, opened 2016-11-03
put back needles for mediacheck PowerPC

michelmno:
related needles were previously removed by commit ids
18aac182d5076f1bd2fb8f44674b809e749c8b08
fec16a146d10e9fd5e98907b03d5f63346993507

but are still required for PowerPC using textmode.

Signed-off-by: Michel Normand <normand@linux.vnet.ibm.com>

PR #128, opened 2016-11-03
put back needle for rescuesystem PowerPC

michelmno:
related needle previously removed by commit id
b5ba3a0a581a18683f20de4cb39763f97094e09c

but still required for PowerPC using textmode


michelmno:
* it needs as prereq the  https://github.com/os-autoinst/os-autoinst-needles-opensuse/pull/127
* this is supposed to solve the failure https://openqa.opensuse.org/tests/296465#step/rescuesystem/3

michelmno:
related question that may need another issue:
* The related test assume the presence of  openSUSE-13.1-gnome64.qcow2 file in /var/lib/openqa/factory/hdd/
* This is OK for the openQA server in  https://openqa.opensuse.org but is not present if installing a brand new openQA server on another machine.
* I manually wget https://openqa.opensuse.org/tests/272415/asset/hdd/openSUSE-13.1-gnome64.qcow2
  But what is the "correct" procedure to be documented in openQA install procedure ?

sysrich:
I believe that to be the 'correct' procedure - the openQA documentation typically assumes you are not trying to replicate 100% exactly what is running in production. The qcow2 is a specific asset for a very specific scenario, the wget of that asset is the way to go..

maybe we could generate new tooling instead of using templates to copy scenarios but to actually copy live scenario configuration AND the assets required by those scenarios from an existing openQA system, but that would be a future feature.

PR #129, opened 2016-11-04
new needle desktop-at-first-boot-minimalx-ppc64

michelmno:
required as workaround for ppc64 because background image
and icons are not correctly displayed tracked at least by bug
https://bugzilla.suse.com/show_bug.cgi?id=929462

This is a workaround needle for gpt, lvm, minimalx tests
that started to fail between 20160927 and 20161010:
https://openqa.opensuse.org/tests/297609#previous
There is no more indication of what was the previously used needle,
this is the reason why I am creating this new needle.


okurz:
thank you. we try to follow a convention for workaround needles so that we know to which bug it refers, please see https://progress.opensuse.org/projects/openqatests/wiki/#applying-quotworkaroundquot-needles

DimStar77:
For reference, the needle that matched before was removed with commit c709defcd6f404efd05c667e5f9ad8dff7d731ba

okurz:
I see, and it wasn't named properly. Maybe a better approach then is to just revert the deletion and update the name.

michelmno:
arg, I forgot the bug number :(

coolo:
and the tags and ... - I actually don't see any reason not to merge this one instead.

michelmno:
Do I need something else for this PR on my side after change made in https://github.com/os-autoinst/os-autoinst-needles-opensuse/pull/129/commits/1d0a169e9fa626d4f9f3df91409673e0a0e61387

PR #131, opened 2016-11-07
Fix missing needles and create new needles

Zaoliang:


okurz:
The commit message could have a bit more details, e.g. what issue is fixed, why this needs to change, to which versions of which product it applies, link to a bug or a fate entry, the choices you made, etc. Also see http://chris.beams.io/posts/git-commit/ as a helpful guide how to write good commit messages. Seems like you are reusing the same git commit message multiple times

Zaoliang:
okurz: can you check the comment again and merge the needles please?

okurz:
`yast2_xinetd_cannot_delete` does not follow our naming convention with the datestamp included

okurz:
merged locally

PR #134, opened 2016-11-08
Fix broken needles

Zaoliang:


coolo:
this doesn't look like fixing, this looks like adding

PR #135, opened 2016-11-08
Update and add needles for yast2_tftp

Zaoliang:


okurz:
IMHO "closed" and "open" port should be turned around but that also would involve the test code and previous needles so I will merge this PR anyway.
@Zaoliang please turn araound "closed" and "open" port in new PRs

PR #136, opened 2016-11-10
Replace trust key option by more relaxed livecd_network_settings-import-untrusted-gpg-key-B88B2FD43DBDC284-20161110

okurz:


coolo:
that's quite a long summary :)

okurz:
On Thursday 10 November 2016 04:09:07 Stephan Kulow wrote:
> that's quite a long summary :)
yes it is

PR #140, opened 2016-11-16
Create missing needles for yast2_tfp

Zaoliang:
http://e13.suse.de/tests/1546#step/yast2_ftp


okurz:
It looks like yast2_ftp right now works for Tumbleweed already: https://openqa.opensuse.org/tests/310759#step/yast2_ftp/49
So I don't know why we should need this PR anymore.
@Zaoliang I am sorry there was no response on this for 12 days by any reviewer.
Please check again if we still need this or not.

okurz:
Also there are some needles which are named "tftp" using a tag "tftp" and I am wondering how this could be correct because the test is for "ftp". Also, the commit message has a typo "tfp" instead of "ftp".

okurz:
no activity, probably outdated anyway

PR #143, opened 2016-12-01
It is the time of the year again

okurz:
christmas needles should never be deleted by admin delete.

Looked up commits deleting needles with CHRISTMAS in name and then reverting
with

```
for i in $(sed -e 's/^commit //g' commits ) ; do git revert --no-commit $i ; done
```


coolo:
please tag them this time, so we can unregister 'CHRISTMAS' in january

coolo:
because admin removal will only remove what is in use

PR #146, opened 2016-12-13
Add needles for https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/2214

okurz:


coolo:
I don't have good memories about this test, but possibly times have changed since then :)

PR #149, opened 2017-01-17
Delete workaround needle without any reference

okurz:
We do not know about any bug. The needle does not adhere to our best practices
and should therefore be deleted. It is easy to be replaced by a proper one
anyway.


okurz:
@lnussel you created that needle, can we delete it? does not look so nice e.g. for update tests which should be green: https://openqa.opensuse.org/tests/336125#step/boot_to_desktop/2

lnussel:
if you delete it the test would fail. But yeah, there should be a bug. it's related to the chaning shade of the start button.

kravciak:
Did you check https://openqa.opensuse.org/tests/336166#step/chrome/20 ?
Might be bsc#1007701

okurz:
ok, deleted. when necessary, we need to create needles with bugref. I will try to keep an eye on opensuse maintenance tests more thoroughly

okurz:
@kravciak The screen you are referencing is the same screenshot.

[bsc#1007701](https://bugzilla.suse.com/show_bug.cgi?id=1007701) is about desktop after upgrade but we are in update tests on a pre-installed image, I doubt it's the same.

PR #154, opened 2017-02-05
Fix misclick on only tab close instead of reader window

okurz:
See video of https://openqa.opensuse.org/tests/348345#step/firefox/7, firefox
is closed instead of reader window.


okurz:
replaced by https://github.com/os-autoinst/os-autoinst-needles-opensuse/commit/3ad03b05c42c4cb64226b85f9a63db0f0bfb9e47

okurz:
@DimStar77 You created the needle I am deleting as you did not use an exclude area and openQA clicks the 'x' on the only open tab instead of the reader window.

PR #156, opened 2017-02-07
Add needles for new Enlightenment version

DimStar77:
Needles going together with https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/2415


DimStar77:
More needles added - now the shutdown also passes cleanly; only 'terminology' missing, but that needs more investigation

PR #158, opened 2017-02-14
Needles for changed flow during toggle home partition

asmorodskyi:
After changes needed for #poo16254 flow of toggle home
partition changed. This needles aim to support this
changes


asmorodskyi:
prove run http://kimball.arch.suse.de/tests/265#step/partitioning_togglehome/3

PR #162, opened 2017-03-06
Reference a workaround needle with boo1025622

michelmno:
as suggested change previously created needle names
with bug https://bugzilla.suse.com/show_bug.cgi?id=1025622

https://github.com/os-autoinst/os-autoinst-needles-opensuse/commit/deb54f48c3e3f0f914bccd6e43ec7ad54cc134af#commitcomment-21146638


okurz:
thank you

PR #171, opened 2017-03-21
Remove wrong named needle

SergioAtSUSE:
I created a new needle in a Tumbleweed test with the name Leap.
I also created a correction with two dashes.
This PR should get rid of that.


SergioAtSUSE:
It works: https://openqa.opensuse.org/tests/373981#step/chromium/20

PR #185, opened 2017-05-02
Adjust the match rate for new desktop selection needles

nilxam:
Adjust the match rate to 100 to role selection since TW now has no default desktop defined, although Leap does, therefore also had 2 needles with "no-default-desktop" tag which just not used in real.


SergioAtSUSE:
LGTM

PR #187, opened 2017-05-05
Fix needles which have been not matched anymore

Zaoliang:
Please see my reference test at:
http://e13.suse.de/tests/2896#


okurz:
did you test against both current openSUSE Tumbleweed and openSUSE Leap?

Zaoliang:
@okurz I tested against TW only

okurz:
Then how do you know if the needles are not matched anymore in other tests? Also, please be more specific with your git commit message.

Zaoliang:
I compared the needles and found that some needles had only matchef for about 75-85 %

okurz:
yes, but you tested only openSUSE Tumbleweed. There is also openSUSE Leap and Maintenance. Or don't we run according tests on all these distributions and versions?
I am not sure if you have permissions to access https://openqa.opensuse.org/admin/needles but this should allow you to check the last match/use of the according needles covering all products currently tested on o3. If you don't have permissions please ask someone with permissions.

Zaoliang:
Please merge this PR. 
Please see ticket:
https://progress.opensuse.org/issues/19022
Without the merged needles it will fail anyway...

PR #188, opened 2017-05-10
Fix yast2_control_center

Zaoliang:
Please see my reference test at: 
http://e13.suse.de/tests/2934#
http://e13.suse.de/tests/2935#


Zaoliang:
Test running  against Leap right now. I will update on this.

okurz:
you are only adding needles here and adjusting one existing one. I am sure this will not break Leap so we can merge it already. Still, testing against Leap makes sense.

PR #191, opened 2017-05-17
Add yast-still-running for dialog shown in foreground (poo#19198)

okurz:


coolo:
isn't this install-shell?

okurz:
hm, good point. It *should* be but the post_fail_hook of y2logstep looks for root-console as fallback when it can't find yast-still-running in http://lord.arch/tests/6259#step/start_install/13 . Should we update yast-still-running instead?

coolo:
most definitely. this screen should hit yast-still-running

okurz:
@coolo updated

PR #202, opened 2017-05-24
Fix yast2_lang.pm for openSUSE Leap

Zaoliang:
- add required needles with tag "ENV-LEAP-1"

Please see my reference test:
http://e13.suse.de/tests/3073#step/yast2_lang

Ticket info:
https://progress.opensuse.org/issues/19354


okurz:
The needles can't work. http://e13.suse.de/tests/3073#step/yast2_lang/11 shows that you only covered the checkbox but not the entry so it could match **any** entry. In http://e13.suse.de/tests/3073#step/yast2_lang/12 you covered a specific entry, in this case "Dutch". This can not work reliably.

Zaoliang:
http://e13.suse.de/tests/3073#step/yast2_lang/12 looks to me normal after I switched to git branch yast2_lang locally.

ducth

PR #203, opened 2017-05-27
Delete invalid needle yast2_lang-yast2-lang_secondary-language-leap-20170524 replaced by new one

okurz:
Selecting only a checkbox without the label is a bad choice as any checkbox
could match.


okurz:
LGTM by asmorodskyi

PR #207, opened 2017-06-07
Poo#18850 firefox java tests for opensuse

rwx788:
In one of the releases java was not supported in firefox. We already have such regression test for sle, which was migrated to TW and Leap.
This commit includes needles required to the test for TW and Leap 42.3 on gnome.

Progress reference: https://progress.opensuse.org/issues/18850
Verification run Leap 42.3: http://gershwin.arch.suse.de/tests/102
Verification run TW: http://gershwin.arch.suse.de/tests/101

Tests PR: https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/3032


okurz:
some of these needles look like a copy from SLES. Maybe you just copied them over from SLES initially. Do you still need them?

kravciak:
I don't know other needles, but I think reader-view you can reuse from SLES - https://openqa.opensuse.org/tests/415683#step/firefox/8

rwx788:
@kravciak somehow reader-view needle didn't work, I believe mainly because on opensuse FF is running in maximized window, while on sle it's not. So I was able to reuse only 2 needles from sle.

rwx788:
@okurz What I did inititially - I copied over all required needles for the test from SLE and replaced most of them with new one's if the didn't match. As stated above, I could reuse only 2 needles from sle. For some needles which visually look same match was giving only 25-30%, so I've decided to recreate.

okurz:
well, it's fine then. LGTM. We can merge, it only affects openSUSE

PR #208, opened 2017-06-12
Add package retry needle as workaround boo1018262

michelmno:
Used in install_and_reboot.pm

Signed-off-by: openQA web UI <openqa@abanc>


coolo:
your webui signs off commits? wow :)

michelmno:
I merged two commits  and set correctly the signoff

michelmno:
@coolo,  is there a specific reason why this PR is not accepted while the one that used it was accepted ? https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/3070

coolo:
I'm innocent on either

PR #212, opened 2017-06-17
Update workaround needles for boo#1025622 to poo#19396

okurz:
https://bugzilla.suse.com/show_bug.cgi?id=1025622 has been closed as invalid,
https://progress.opensuse.org/issues/19396 is the follow-up.


okurz:
*bump*

SergioAtSUSE:
I need a verification run ;)

nicksinger:
Filenames between json and png match so LGTM. But I agree with @SergioAtSUSE just for the sake of safety.

okurz:
`;)` to me means you don't actually need a verification run so I am reading @nicksinger's LGTM

PR #213, opened 2017-06-19
Verify intallation overview list is loaded

rwx788:
During execution on slow system loading of overview may take time,
whereas we start pressing tab key until get to the software section. It
is in the top, and if we press tab couple of times before, we won't be
able to reach software section in 20 presses for some installations.
Hence, simply add wait_screen check to avoid this kind of failures.

See poo#12322.


asmorodskyi:
@okurz please take a look

PR #222, opened 2017-07-04
Add needle for opaque KDE logout screen

StefanBruens:
At least on aarch64, the logout screen seems to have no transparency,
workaround for poo#20258.


okurz:
please keep the date in the name at the end

okurz:
@StefanBruens thanks for your contribution. Little hint: There is no notification sent by github when there is just updates to commits but no comment. So I recommend to always write a comment if you want someone to take another look into the PR, e.g. just "updated" suffices.

StefanBruens:
@okurz Thx, noted.

PR #223, opened 2017-07-05
Create missing needles for yast2_lan_restart.pm

Zaoliang:
- add yast2_lan_restart back to extratest/yast2_gui test for openSUSE TW/LEAP
- please see the related PR: https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/3236

Please my reference tests:
openSUSE TW - http://e13.suse.de/tests/3259
openSUSE Leap - http://e13.suse.de/tests/3252


okurz:
The test PR got merged but not the needles which broke the tests -> https://openqa.opensuse.org/tests/441902#step/yast2_lan_restart/16 I merged the needles now even though they could have been improved. Please keep in mind for the next time to reference the needles PR in the description of the according test PR as well. Also, I think the needles could be separated by ENV-LEAP-0/1 but as they are very specific to the yast2_lan_restart I guess we are fine to just check two needles in each step for both TW and Leap so good enough for now.

PR #226, opened 2017-07-07
Cleanup installation overview needles

rwx788:
We have created new needles to verify release notes button on
installation settings overview screen. It also matches other newly
created needles without button. Hence as a cleanup, we remove them to
rely only on needles with release notes button area. This removes
identified duplicates, which won't be removed automatically, as both
match.

Examples of such duplicate needles:
https://openqa.opensuse.org/tests/442300#step/installation_overview/1
https://openqa.opensuse.org/tests/442307#step/installation_overview/1
https://openqa.opensuse.org/tests/442301#step/installation_overview/1
https://openqa.opensuse.org/tests/442305#step/installation_overview/1
https://openqa.opensuse.org/tests/442253#step/installation_overview/1


SergioAtSUSE:
I need a verification run ;)

rwx788:
There are 5 runs above, do you really want me to get of them? =)

SergioAtSUSE:
But, those are not verification runs, those are example of the duplicates, aren't they?. :confused: 

Also, I am seeing on the first example https://openqa.opensuse.org/tests/442300#step/installation_overview/1 that it is matching one needle that is deleted on this PR `inst-overview-textmode-20160831`.

A verification run would be one where there is not matching one of the deleted needles.
And also I expect the verification run to not be on openqa.opensuse.org, since this might mean that the changes are already on productive instances.

SergioAtSUSE:
As spoken with @rwx788 
We will wait for tomorrow for that PR to also removed the next ones duplicates.

okurz:
@rwx788 merged, now please monitor o3 jobs closely. thanks.

rwx788:
@okurz Will do so.

lnussel:
this breaks staging.

why does this now check for a release notes button anways? if we wanted to verify that release notes work we should have a separate test for that.

rwx788:
@lnussel My bad, terribly sorry for that. Previously we had good_buttons test which aim was to verify release_notes button as well, but at some point needle without it was created and test became useless. For SLE we test release notes separately, but here as part of improvement we wanted to introduce new needles. Could you please point me to the staging job group you are referring to?

lnussel:
https://openqa.opensuse.org/group_overview/2
https://openqa.opensuse.org/group_overview/6

lnussel:
I've reverted the commit so we can continue

nilxam:
Note that, for openSUSE, we don't have the release note button in the Staging testdvd.

rwx788:
Was not aware of it. Then yes, let's keep it as it is after revert and if required, we add separate test for release notes. Thanks!

PR #230, opened 2017-07-14
Add firefox_clean needles

ostruzina:
Related to https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/3243.

PR#3243 above adds `firefox_check_default` and `firefox_check_popups` methods into the x11regressiontest library, so they can be reused in various test modules testing firefox.

These methods will be then used in the x11/firefox module.

To make the subroutines more generic and usable in different scenarios, I have used a new needle tag `firefox_clean`, which is used to check that firefox has no pop up windows left open.

Verification run:
Leap 42.3: http://dreamyhamster.suse.cz/tests/384#step/firefox/4
Tumbleweed: http://dreamyhamster.suse.cz/tests/383#step/firefox/4


ostruzina:
Apologies, I messed up my git history; will submit a new PR.

PR #237, opened 2017-07-19
Add needles for hwsim_wifi test

okurz:


okurz:
See https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/3313 for the corresponding test adaptions and enablement in the test plan.

@nicksinger: For now the openSUSE Leap 42.3 test stops early because the wifi can not be discovered from within the gnome widget. I suspect hostapd not properly running the network: http://lord.arch/tests/7016#step/mac80211_hwsim/68
Ideas?

okurz:
updated, please merge if ok

okurz:
LGTM by @Soulofdestiny

PR #241, opened 2017-07-20
New needles for http yast test

rwx788:
These needles aim making test more stable. In some runs particular
screens get distorted which prevents test from passing. Whereas from
functional perspective, this is not any kind of issue.


okurz:
s/particuar/particular/

okurz:
approved after typo fix

rwx788:
Fixed in commit message and in description here. Thanks!

PR #245, opened 2017-07-21
Needles for stabilizing yast2_ui tests

rwx788:
Now I've changed approach to intoroducing changes to multiple tests in
one run according to my findings on the runs. Process reminds a bit
pluging holes in a sinking boat with fingers. On top of that chances to
spot new potential issue are higher when tests are executed as a suite
and not separately.
See poo#19922 and all blocking tasks.


consoletest_finish-gnome-TW-20170721.json
```diff
@@ -0,0 +1,27 @@
...
+  ],
+  "tags": [
+    "x11",
+    "generic-desktop",
+    "reboot-after-installation",
+    "ENV-DESKTOP-gnome",
+    "ENV-DISTRI-opensuse",
+    "ENV-INSTLANG-en_US"
+  ]
+}
```
okurz: hm, what's different about this needle? Why do we need another one?

yast2_proxy_http_refresh_patterns_selected-20170721.json
```diff
@@ -0,0 +1,15 @@
...
+  "area": [
+    {
+      "ypos": 82,
+      "height": 11,
+      "type": "match",
+      "width": 141,
+      "xpos": 40
+    }
+  ]
+}
```
okurz: aren't there actually two entries selected in the screenshot?

consoletest_finish-gnome-TW-20170721.json
rwx788: It's because of updates pop-up. Even though we disable package-kit service.

yast2_proxy_http_refresh_patterns_selected-20170721.json
rwx788: They are and it happens once in while cause of ncurses gui lags, so new needle has smaller area to match and such scenario won't fail the test.

yast2_ftp_create_syntax_error_bsc#1041829.json
```diff
@@ -0,0 +1,15 @@
...
+      "ypos": 330,
+      "width": 307,
+      "height": 91,
+      "type": "match"
+    }
+  ],
+  "tags": [
+    "yast2_ftp_syntax_error_bsc#1041829"
+  ]
+}
```
okurz: workaround property missing. Also missing datetime in filename

consoletest_finish-gnome-TW-20170721.json
okurz: But there are already many needles with the popup excluded. Maybe you should delete some that do not have it at all?

rwx788: So there were 2 needles which do the job, I've removed old one and introduced new one with update are excluded, as well as time.

yast2_ftp_create_syntax_error_bsc#1041829.json
rwx788: Fixed

PR #249, opened 2017-08-02
Add required needles for extending yast2_software_management.pm

Zaoliang:
- add checks for search, pattern, show list and summary
- please also see ticket for details:
  https://progress.opensuse.org/issues/20966
- please also check PR: 
  https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/3371
- reference test: 
   http://e13.suse.de/tests/3919#step/yast2_software_management


okurz:
I strongly recommend not to cover huge groups of variable content, e.g. as in http://e13.suse.de/tests/3919#step/yast2_software_management/15 . This could be very annoying to maintain whenever content changes and I don't see value in that from testing perspective

Zaoliang:
@okurz will remove RPM group then

Zaoliang:
successful test after needles of "rpm group" got removed:
http://e13.suse.de/tests/3921#step/yast2_software_management/

okurz:
Please check all needles not only the rpm group one for bigger areas covered where the content could change. Also, it's recommended to not cover more than one column per needle area, e.g. as in https://github.com/os-autoinst/os-autoinst-needles-opensuse/pull/249/files#diff-35fb331d80a4003503342587fc9a7148

Zaoliang:
@okurz I've changed now some big size to smaller and reasonable size.

Zaoliang:
new reference test:
http://e13.suse.de/tests/3948#step/yast2_software_management

PR #251, opened 2017-08-06
inkscape-1-kde-20170806 for opensuse-Tumbleweed-DVD-aarch64-Build2017…

StefanBruens:
…0804-kde@aarch64


StefanBruens:
see poo#21844

foursixnine:
@StefanBruens I opened this one today: https://bugzilla.opensuse.org/show_bug.cgi?id=1053378 I'd say that as a workaround this would help, but there's definitely a problem

StefanBruens:
@foursixnine 
Both renderings (x86_64 and aarch64) are actually correct, just "different". It may be caused by different DPI settings or some other driver peculiarity.

Both LXDE and GNOME already have added needles to deal with this difference, while the KDE openQA run always fails.

PR #253, opened 2017-08-18
Needles required for the cryptlvm with storage-ng

ancorgs:
I did run `grep` in the `autoinst-log.txt` of this successful test (only accessible at SUSE): https://yast-openqa.suse.cz/tests/13400#step/partitioning_lvm/2

These are the needles that were used to succeed but that are not present in upstream openQA.

I fact, I intentionally ignored three needles called `first_boot-generic-desktop-20170708`, `role-selection-20170324` and `role-selection-leap-20170706` that I assume to be workarounds, since they don't look related to storage. So I indeed included here the very minimum I assume is needed.


ancorgs:
Hmmm, I just realized this includes some console needles. So obviously I didn't filter enough. Let me improve it

PR #254, opened 2017-08-18
Needles required for the cryptlvm with storage-ng

ancorgs:
This is a refined version of https://github.com/os-autoinst/os-autoinst-needles-opensuse/pull/253 with a minimum set of needles than I'm pretty sure will be needed


coolo:
note we need that for sle too

PR #255, opened 2017-08-23
Add tty6 selected needles

rwx788:
For many distris we switched to tty6 as root-console. Hence require a
new needle for that.


tty-6-selected-20170823.json
```diff
@@ -0,0 +1,23 @@
+{
+  "tags": [
+    "ENV-15ORLATER-1",
```
okurz: doesn't make sense for openSUSE

rwx788: Right, seems that I've started adding it automatically...

PR #257, opened 2017-08-25
Remove needle from yast2_xinetd

Zaoliang:
-  need to remove needle created directly on openQA
   because I've implemented softfail in yast2_xinetd.pn now
-  please see my PR:
   https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/3464


okurz:
I don't see a benefit in the test code change but the name of the workaround needle should be adjusted to our convention, see https://progress.opensuse.org/projects/openqatests/wiki/Wiki#applying-quotworkaroundquot-needles

PR #261, opened 2017-09-01
Add gnome-control-center 3.26 needles (reworked layout/workflow)

DimStar77:


DimStar77:
Needles matching https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/3510

gnome_control_center-started-20170831.json
```diff
@@ -16,7 +16,8 @@
     }
   ],
   "tags": [
-    "gnome-control-center-started"
+    "gnome-control-center-started",
+    "gnome-control-center-new-layout"
```
asmorodskyi: I think 'new' is not good word here , because you will get into a trouble when newer will appear :) 
suggest to change it to version number ? "gnome-control-center-3-26-layout" ?

DimStar77: and in 6 month you want to tag GNOME 3.28 control-center with a 3.26-layout needle? That will be even more confusing imho - and chances are low that the entire shell will be switched.

Other/better alternatives?

asmorodskyi: no better alternatives , so you can leave it as it is

PR #265, opened 2017-09-12
Add needles for testsuite regression-gnome

chuchingkai:
see also: poo#23884

Validation test: http://10.67.17.30/tests/1908


chuchingkai:
PR for test case: https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/3558

PR #266, opened 2017-09-18
Delete soft-failed needles for fixed bug replaced by inst-welcome-text-20170918

okurz:


SergioAtSUSE:
Some progress ticket or verification run?

okurz:
there is no progress ticket. I am deleting workaround needles which relate to a fixed bug. The jobs have inst-welcome-text-20170918 instead to match on

PR #268, opened 2017-09-28
Delete misleading needle yast2_xinetd_startup-20170824 covering a "core dump"

okurz:
The needle was not marked as soft-fail needle linked to a bug but trying to
match a "core dump" message which was not even properly handled in the test
code which is looking for the "core dump" tag which was also not set on this
needle.


okurz:
@zluo

DimStar77:
👍

PR #274, opened 2017-10-10
Prevent gnucash needle matching when main window is not yet ready

okurz:
Fixes problems as observed in
openqa.opensuse.org/tests/501767/modules/gnucash/steps/45

Previously x11_start_program was waiting way longer before checking. The older
needles only looked for the "tip of the day" popup but press "ctrl-h" to open
the help which is only working when the main window is responsive. Because
x11_start_program is way faster now the main window is not yet responding to
key presses and the test fails to find the help window.

Verification run: http://lord.arch/tests/7667

Related progress issue: https://progress.opensuse.org/issues/25876


DimStar77:
awesome! Thanks for debugging that and finding the right bullets

PR #283, opened 2017-10-26
Fix autoyast_reboot due to disable timeout

jknphy:


reboot-bios-seabios-20171026.json
```diff
@@ -0,0 +1,17 @@
...
+      "width": 144,
+      "ypos": 1,
+      "height": 20,
+      "xpos": 1,
+      "type": "match"
+    }
+  ],
+  "tags": [
+    "bios-boot",
+    "ENV-DISTRI-sle",
```
okurz: that doesn't make sense. SLE needles are not stored in this repo

jknphy:
Fixed. bios needle still needed for removing version specific reducing the rectangle of the current one in this repo with similar name.

PR #286, opened 2017-11-21
New needles for partitioning and yast errors

cwh42:


partitioning_expert-20171011.json
```diff
@@ -0,0 +1,15 @@
+{
+  "tags": [
+    "really-delete"
```
okurz: the file name should have this included. Also I recommend to be a bit more explicit here with the tag name, e.g. "partitioning_expert-really-delete". Then the filename could be correspondingly "partitioning_expert-really-delete-20171011.json"

okurz:
please change the author of the git commit to your real user, not "openQA webUI user"

PR #287, opened 2017-11-21
Needles for yast control center test module on leap 15

rwx788:
For leap 15 we required new needles for storage ng and udapted needle
for hypervisor (due to changed tag name).

See [poo#25672](https://progress.opensuse.org/issues/25672).


SergioAtSUSE:
Could you provide a verification run?

okurz:
In commit message and PR description s/udapted/updated/

rwx788:
@okurz fixed, thanks!

PR #295, opened 2017-12-06
Fix upgrade_select_opensuse

DrMullings:
Needles that allow completion of upgrade_select_opensuse for
Leap 15

PR with code: https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/4016


okurz:
Please avoid the very uncommon capitalization in the git commit message and PRs subject line.

I don't see a reason for a version-specific needle. Please do not cover over the "15.0" part. See https://progress.opensuse.org/projects/openqatests/wiki/#don39t-include-version-specific-content-in-needles for reasoning

DrMullings:
Since this is an update test shouldnt we check the version?

okurz:
Hm, maybe. Then at best create *both* variants. Once with version and once without.

DrMullings:
Then we have one that always matches, I am not sure if this is wanted here?
The needle itself is not needed to Fix the double beta warning, it just makes the test pass at the end, so we could handle this as two items?

okurz:
Again: Please avoid the very uncommon capitalization in the git commit message and PRs subject line.

Disregard the part about version in the needle

DrMullings:
Adapted

PR #300, opened 2017-12-15
Add needles for leap 15 and TW to disable online repos

rwx788:
See [poo#17436](https://progress.opensuse.org/issues/17436).


asmorodskyi:
probably you should add tag which will filter out this needles for earlier versions ?

rwx788:
They are not version specific, so should work with older versions as well, especially for text mode...

PR #311, opened 2018-01-24
Add some needles needed for yast2_lan_restart in leap15

jknphy:
https://progress.opensuse.org/issues/28666
Adding some needles needed, but still need more work, but at least the test can move forward a little bit.


yast2_lan_restart-yast2_lan_global_options_tab-leap-15-20180124.json
```diff
@@ -0,0 +1,15 @@
...
+    {
+      "ypos": 144,
+      "xpos": 85,
+      "type": "match",
+      "width": 137,
+      "height": 29
+    }
+  ],
+  "tags": [
+    "yast2_lan_global_options_tab"
```
rwx788: I guess if this is leap specific, let's add ENV-LEAP-1 to all needles here, or to none if we expect them to work on TW, for example.

jknphy: yes, these are leap specific needles. Added.

jknphy:
Created some more needles for this tests in order to not leave it half tested, but I found that  in this screen https://github.com/os-autoinst/os-autoinst-needles-opensuse/pull/311/files#diff-c25bbbc4d622d9e7d4b401d964706858 we cannot select a line in the list, maybe a new visualization for tables or we will need to adapt the test in future.

PR #319, opened 2018-02-08
Needle to assert that firefox is started correctly

rwx788:
On KDE we have instability with desktop runner and for some reason
beginning of the line sometimes is not there.
To work this around, we will retry starting application if needle
doesn't match.

[code](https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/4381)

See [poo#25972](https://progress.opensuse.org/issues/25972).


rwx788:
LGTMed by @okurz

PR #324, opened 2018-02-14
Needles to select root partition in splitusr test suite

rwx788:
Fixes https://openqa.opensuse.org/tests/609742

Note: vda-selected somehow didn't match with same needle on TW and Leap, as area cannot be small enough, so that's the reason for 2 needles

[Code here](https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/4413)


okurz:
spliturt?

rwx788:
Fixed

PR #331, opened 2018-02-27
Delete invalid ext4 needles

rwx788:
We have needles where matching area is too big and ext2 fs matches when
we expect ext4. Removing those. Thanks to @okurz finding it, as test
doesn't fail.

[Verification run](http://g226.suse.de/tests/868#).

See [poo#32362](https://progress.opensuse.org/issues/32362).


nicksinger:
LGTM

PR #332, opened 2018-02-27
More wrong needles that match ext4 due to big area

rwx788:
[PR#331](https://github.com/os-autoinst/os-autoinst-needles-opensuse/pull/331) wasn't enough.

See [poo#32362](https://progress.opensuse.org/issues/32362).


nicksinger:
I wonder if we face these issues again soon. The two needles you delete there are actually from a different screen. How can we make sure $somebody will not create them again (with a different screen)?

To resolve the initial issue it looks good to me but as already phrased; we need to think about a long-term solution.

PR #338, opened 2018-03-20
Needle for partitioning lvm test, which fails in staging

rwx788:
[Code can be found here](https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/4666).


okurz:
conflicting file. I assume someone created the needle already on o3. Do we still need this PR?

rwx788:
Yeah, this is not required.

PR #339, opened 2018-03-21
Add yast2 snapper ncurse needles for openSUSE

mitiao:
See: https://progress.opensuse.org/issues/33520
Verification runs:
[Leap15](http://openqa-apac1.suse.de/tests/517#step/yast2_snapper_ncurses/37)
[Tumbleweed](http://openqa-apac1.suse.de/tests/518#step/yast2_snapper_ncurses/37)


okurz:
Thanks for a fast response and your fixes

PR #340, opened 2018-03-26
Add needles of gdm to fix poo#33484

chuchingkai:
These needles are created by crop.py, based on needles of testsuite gnome-gdm on o.o.o:
- Leap 15: https://openqa.opensuse.org/tests/640083#step/first_boot/3
- TW: https://openqa.opensuse.org/tests/642729#step/first_boot/3

The default user of gdm will be selected for the scenario of reboot or first boot, we can just press enter to login. But for the scenario of log out and re-login we have to use assert_and_click to work around
bsc#1086425.

These needles will be used to check if the keyboard focus of gdm user is lost.

  - see also: [poo#33484](https://progress.opensuse.org/issues/33484), [bsc#1086425](https://bugzilla.suse.com/show_bug.cgi?id=1086425#add_comment)
  - related PR: https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/4689


chuchingkai:
The fonts in Leap and TW are different

PR #344, opened 2018-03-30
Delete obsolete .json file for not-existant yast2_bootloader-yast2-bootloader_defautl-boot-section_tw-20170615.png

okurz:


rwx788:
I wonder how did it work, as usually openQA fails in case of missing needles.

PR #346, opened 2018-04-01
Add needles for wine test

okurz:


steam-login-startup-libgl_error-20180331.json
```diff
@@ -0,0 +1,15 @@
+{
+  "tags": [
+    "steam-startup-libgl_error"
```
asmorodskyi: Can't find usage of this tag in https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/4749

steam-agreement-ok-leap-20180401.json
```diff
@@ -0,0 +1,24 @@
+{
```
asmorodskyi: Why you need this needle ? Looks identical to steam-agreement-ok but with console in background which should be unrelated.

steam-login-startup-libgl_error-20180331.json
okurz: See https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/4748/files#diff-32797c94ffb120abddf9f3997212c570R29

steam-agreement-ok-leap-20180401.json
okurz: The PR includes two variants, for openSUSE Tumbleweed, no prefix in name, and openSUSE Leap, with "-leap-" in name

steam-login-startup-libgl_error-20180331.json
asmorodskyi: ok

PR #347, opened 2018-04-03
Remove zypper_ar screen check

rwx788:
See
[PR#4679](https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/4679),
we removed only usage of the needle there, hence not required anymore.


okurz:
In this case I strongly recommend to merge the needle PR *after* the test code change as it's about removal of needles, not adding

rwx788:
Code is merged, needles are not used. @okurz, are you fine if we merge this PR?

PR #348, opened 2018-04-04
[WIP] Add network-not-configured needle for offline mode

asdil12:


okurz:
I guess you can simply remove the [WIP] and we can merge these already as more needles shouldn't break tests even though we can only merge the test code changes after the os-autoinst change was deployed

PR #351, opened 2018-04-11
Rename log view needle for tftp test

rwx788:
As per comment by @okurz in
[poo#34072](https://progress.opensuse.org/issues/34072).


okurz:
seems you have referenced the wrong ticket. Should be https://progress.opensuse.org/issues/34072 I guess

okurz:
and why would I ask to rename yast2_tftp_view_log_show-20161108.json to yast2_tftp-view_log_show-20180411.json ?

The ticket stated
* Delete needle "yast2_tftp_view_log_show-20161108" which can match prematurely before the actual error popup shows up
* The needle "yast2_tftp_view_log_error" seems to be the correct one to match but does not follow our naming convention -> rename

now you got me confused

rwx788:
@okurz ah, I see, I read it by diagonal, so thought log_show is actually correct. I'll rename error then too. I can explain why removing a needle won't help if you're interested. I've renamed error needles as well.

okurz:
Please delete the file "yast2_tftp-view_log-error.json/png" as it does not have a timestamp

rwx788:
Removed. Thanks!

rwx788:
@okurz any objections to merge?

PR #354, opened 2018-04-18
Excluded notification area for leap pop-up

mloviska:
Affected test module keymap_or_local in gnome desktop.

- Verification run [opensuse-15.0-NET-x86_64-Build219.1-update_Leap_42.2_cryptlvm@uefi](http://dhcp151.suse.cz/tests/2188#step/keymap_or_locale/10)
- Ticket [[functional][u][fast][medium] test fails in keymap_or_locale - potentially lost focus due to popup?](https://progress.opensuse.org/issues/34321)

- https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/4948


mloviska:
I am not so confident, that it helps. [Evidence](http://dhcp151.suse.cz/tests/2169#step/keymap_or_locale/5)

kravciak:
I think problem here is that you are disabling notifications for root (using root-console). Try disabling them from user-console = for current user.

mloviska:
nope, no hero.
http://dhcp151.suse.cz/tests/2187#step/keymap_or_locale/5

kravciak:
No hero = no more ideas :( ..sorry

consoletest_finish-gnome-leap15.0-20180416.json
```diff
@@ -0,0 +1,35 @@
+{
+  "tags": [
+    "ENV-DESKTOP-gnome",
+    "ENV-DISTRI-opensuse",
+    "ENV-INSTLANG-en_US",
+    "ENV-VERSION-15.0",
```
okurz: please avoid version specific needles. This would not match for 15.1 and I think we should try to match it. For SLE we have `ENV-15ORLATER-1`. I think we do not have it for openSUSE, I think, so better just delete this line

okurz:
We don't need this anymore, do we?

mloviska:
There is still chance that some kind of pop up can appear at very beginning of the test, because the system has booted to x11 in first_boot test already. This kind of situation occurs very rarely, I thought it might be useful to keep this needle.
e.g.
http://dhcp151.suse.cz/tests/1803#step/keymap_or_locale/1

PR #362, opened 2018-05-04
Delete obsolete soft-fail needle without bugref

okurz:
I stumbled over the needle which has the "workaround" property but does not
link to a bug.

we try to follow the convention as described on
https://progress.opensuse.org/projects/openqatests/wiki/Wiki#applying-quotworkaroundquot-needles
so the needle name should have a reference to a product bug on bugzilla - or in
rare exceptions - to progress tickets.

The original issue that the needle tried to detect seems to be fixed meanwhile,
upstream bug https://bugs.kde.org/show_bug.cgi?id=382668 is.


okurz:
deletion approved by @StefanBruens

PR #367, opened 2018-05-10
Test the Screen Sharing in GNOME e.g. the vino package

pdostal:
This is related to https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/5015


okurz:
In the filenames please keep the timestamp at the end. See existing needles for an example

asmorodskyi:
you need to keep in mind that needles which you creating will be used not in isolation for your test exclusively but will be added to dozen other needles. So when you want to click on GNOME settings to get to vino settings it is not correct to call this needle vino-settings because it is too specific. 
Also since you switch from your custom way of doing logout you can drop needles with tags - "log-out" and "really-logout"

pdostal:
@okurz timestamps are still part of the filename.

@asmorodskyi removed. Is there a way how to find a needle effectively?

okurz:
Please rebase, all the "merges" are hard to track and confuse reviewing

okurz:
I see names like vino-sharing-20180510_1, so please keep the timestamp *at the end*, meaning s/vino-sharing-20180510_1/vino-sharing-1-20180510/

pdostal:
@okurz in this case the [vino-sharing-20180510.png](https://github.com/pdostal/os-autoinst-needles-opensuse/blob/cdf3c405d2fafe52ccaecbca191cadd9340f98e1/vino-sharing-20180510.png) and [vino-sharing-20180510_1.png](https://github.com/pdostal/os-autoinst-needles-opensuse/blob/cdf3c405d2fafe52ccaecbca191cadd9340f98e1/vino-sharing-20180510_1.png) are not the same. How to differ them?

okurz:
Yes, this is why I proposed the "-1-" as part of the filename, but *before* the timestamp

asmorodskyi:
@pdostal your PR still contains 58 commits

pdostal:
The PR is now clean.

pdostal:
Should I ask create a PR for the SLE needles in the other repository?

pdostal:
Squeezed & rebased.

pdostal:
@rwx788 I'd like to create the PR for SLE needle by myself as well as soon as the server will be up.

rwx788:
@pdostal if you have access to the, it will be very appreciated! gitlab should be up by now. Otherwise, let me know and we'll make it happen. Cheers!

pdostal:
The MR 868 with needles for SLE was created. Cheers! :)

PR #369, opened 2018-05-12
new snap-default needle

michelmno:
as required by already committed change
https://github.com/os-autoinst/os-autoinst-distri-opensuse/commit/a5c75efdbf5fa62cc04ba31b598d95a65be29ae1


michelmno:
There was no "snap-default" so the PowerPC  test is failing without it on o3:
https://openqa.opensuse.org/tests/674539#step/grub_test/7

okurz:
Sorry about that if it caused any inconvenience. So far I do not know a better way to ensure necessary needles are created in before or very soon after merging, suggestions welcome :)

PR #371, opened 2018-05-16
Fix problem with pressing play in vlc.pm

Zaoliang:
- see poo#33523 and 16516
- create a needle for match before pressing play button
- verification run:
  http://e13.suse.de/tests/2429#step/vlc


kravciak:
String `/data/Big_Buck_Bunny_8_seconds_bird_clip.ogv` is not typed properly, how does needle change help?

Zaoliang:
@kravciak the problem before pressing play is no check for correct input or too early.
So assert_screen forces it to check needle match. In this case it helps.

Zaoliang:
@kravciak I created one needle on o3 directly and then I found that needle's position is not good because it is not same place when it tries to match. so that is why I create a new one which is work better for match. Important here is to force to check url input.

kravciak:
Ok then, let's see if it helps

okurz:
Looks like a good idea

PR #373, opened 2018-05-25
Add xfce_lightdm_logout_login for Leap 15.0 for aarch64

ggardet:
This will fix https://openqa.opensuse.org/tests/682504#step/xfce_lightdm_logout_login/5


xfce_lightdm_logout_login-leap15.0.json
```diff
@@ -0,0 +1,30 @@
+{
```
okurz: Please keep the timestamp in the filename of the needle as we have as an convention for - I hope - all of our needles. Unfortunately we do not yet have a check in this needle repo for this. So e.g. name it "xfce_lightdm_logout_login-leap15.0-20180525.json" and the .png accordingly

ggardet:
Not sure why I did not add Timestamp as usual. 
Anyway, the problem went away so I close this no more needed PR.

PR #376, opened 2018-05-29
Match second drive for partitioning_firstdisk

mloviska:
Whenever a hdd is used as an installation medium, it should not be offered as a storage for new OS. In other words partitioner does not propose installation medium as a suitable storage for deployment of OS.
Partitioner skips this [hdd offer](https://openqa.opensuse.org/tests/678603#step/partitioning_firstdisk/1) in case there are used only 2 hdds and of them is installation medium.

OpenQA test variable NUMDISKS should be updated to use 3 hdds.

* Related ticket: [[functional][y] test fails in partitioning_firstdisk of "external_iso"](https://progress.opensuse.org/issues/36363)
* Verification run: [opensuse-Tumbleweed-DVD-x86_64-BuildSnapshot20180527-external_iso@64bit](http://dhcp151.suse.cz/tests/3221#step/partitioning_firstdisk/2)
* Old storage stack is not affected with additional drive
[sle-12-SP4-Server-DVD-x86_64-Build0246-external_iso@64bit](http://dhcp151.suse.cz/tests/3273#step/partitioning_firstdisk/1)


okurz:
yeah, that's sneaky ;)

mloviska:
Yeap, it is. :D However we do not need this needle anyway.

mloviska:
I am not really sure what we can do with it. But basically yes, we can.

okurz:
OK, I did. You can also close PRs on your own.

PR #377, opened 2018-05-29
Add README for the project

rwx788:
See [poo#28654](https://progress.opensuse.org/issues/28654).


okurz:
A very good start. Thanks. We can still add some "the" in a later step ;)

PR #390, opened 2018-07-04
Add inital self-test for travis

nicksinger:
This adds a simple self-test for this repo executed on travis-CI.
A first impression how it looks can be seen on my personal branch: https://travis-ci.org/nicksinger/os-autoinst-needles-opensuse

### DIFF
--- a/tests/test.py
+++ b/tests/test.py
@@ -1,0 +1,48 @@
+#!/usr/bin/python3
+
+import os
+import sys
+import re
+import json
+
+scriptdir = os.path.dirname(os.path.realpath(__file__))
+needledir = os.path.join(scriptdir, "..")
+
+returncode = 0
+
+def error(msg):
+    global returncode
+    returncode = 1
+    print(msg, file=sys.stderr)
+
+# Get all needles
+needles = set()
+for f in os.listdir(needledir):
+    if f.endswith('.json') or f.endswith('.png'):
+        needle, ext = f.rsplit('.', 1)
+        needles.add(needle)
+
+# Check for missing parts
+for needle in sorted(needles):
+    jsonfile = os.path.join(needledir, needle + '.json')
+    pngfile = os.path.join(needledir, needle + '.png')
+
+    # Check for file existence
+    if not os.path.isfile(jsonfile):
+        error("Needle '{}' is missing its JSON file!".format(needle))
+        continue # parsing the json makes no sense
+
+    if not os.path.isfile(pngfile):
+        error("Needle '{}' is missing its PNG file!".format(needle))
+
+    # Check JSON content
+    n = None
+    with open(jsonfile) as f:
+        n = json.load(f)
+
+    # Check if workaround tag exists if bugref is in name
+    if re.search(r'-(poo|bsc|bnc|boo)[0-9]+-', needle):
+        if not 'workaround' in n['properties']:
+            error("Needle '{}' has bugid in name but no workaround tag set!".format(needle))
+
+sys.exit(returncode)
--- a/tests/test.py
+++ b/tests/test.py
@@ -36,13 +36,12 @@ for needle in sorted(needles):
         error("Needle '{}' is missing its PNG file!".format(needle))
 
     # Check JSON content
-    n = None
+    n = {}
     with open(jsonfile) as f:
         n = json.load(f)
 
     # Check if workaround tag exists if bugref is in name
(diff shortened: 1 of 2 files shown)


okurz:
so I like the idea in general.

Would you be ok to add a Makefile to call the tests?

And I have a problem with your python script and the indentation of 2 spaces :/ And on top. When you want to add python code then you should add checks for that python code as well. Yes, I'm serious :) E.g. how about reusing the parts from openqa_review checks for that?

tests/test.py
```diff
@@ -0,0 +1,48 @@
...
+
+scriptdir = os.path.dirname(os.path.realpath(__file__))
+needledir = os.path.join(scriptdir, "..")
+
+returncode = 0
+
+def error(msg):
+  global returncode
+  returncode = 1
+  print(msg, file=sys.stderr)
```
okurz: I guess we could just go for pytest code. That should simplify a lot and we have the whole functionality of test runners, nice output, etc.

tests/test.py
```diff
@@ -0,0 +1,48 @@
...
+
+  # Check JSON content
+  n = None
+  with open(jsonfile) as f:
+    n = json.load(f)
+
+  # Check if workaround tag exists if bugref is in name
+  if re.search(r'-(poo|bsc|bnc|boo)[0-9]+-', needle):
+    if not 'workaround' in n['properties']:
+      error("Needle '{}' has bugid in name but no workaround tag set!".format(needle))
```
okurz: Actually I would allow to mention a bug in the name without a workaround property. Might be a "RESOLVED WONTFIX" bug for reference. However, we should disallow the other way around, workaround property must always be accompanied by a bugref in the name.

tests/test.py
```diff
@@ -0,0 +1,48 @@
...
+  if not os.path.isfile(pngfile):
+    error("Needle '{}' is missing its PNG file!".format(needle))
+
+  # Check JSON content
+  n = None
+  with open(jsonfile) as f:
+    n = json.load(f)
+
+  # Check if workaround tag exists if bugref is in name
+  if re.search(r'-(poo|bsc|bnc|boo)[0-9]+-', needle):
```
okurz: I suggest to turn any reference to 'bnc[0-9]' into an explicit fail by itself to not allow any of these old outdated references

tests/test.py
asdil12: Something like this?

```diff
diff --git a/tests/test.py b/tests/test.py
index 518449b5..28129658 100755
--- a/tests/test.py
+++ b/tests/test.py
@@ -41,8 +41,10 @@ for needle in sorted(needles):
     n = json.load(f)
 
   # Check if workaround tag exists if bugref is in name
-  if re.search(r'-(poo|bsc|bnc|boo)[0-9]+-', needle):
-    if not 'workaround' in n['properties']:
-      error("Needle '{}' has bugid in name but no workaround tag set!".format(needle))
+  needle_has_bugref = bool(re.search(r'-(poo|bsc|bnc|boo)[0-9]+-', needle))
+  needle_has_workaround = 'workaround' in n.get('properties', [])
+
+  if needle_has_workaround and not needle_has_bugref:
+    error("Needle '{}' has workaround tag but no bugid in name!".format(needle))
 
 sys.exit(returncode)
```

nicksinger:
> so I like the idea in general.
> 
> Would you be ok to add a Makefile to call the tests?

Yes, @asdil12 mentioned it and I decided actively against this. IMHO is is just another dependency which does not add any value. Even the opposite happens: people may get confused why there is a Makefile for a repo only containing json and png - "Is there something to compile?".

> And I have a problem with your python script and the indentation of 2 spaces :/ And on top.
Even though with tabs, one could easily set his editor to the desired indent; I've changed it ;)

> When you want to add python code then you should add checks for that python code as well. Yes, I'm serious :) E.g. how about reusing the parts from openqa_review checks for that?
And who tests the test of the test?

tests/test.py
nicksinger: Definitely a thing that could be considered in in the future

asdil12:
LGTM.

tests/test.py
nicksinger: Why should a needle be invalid if the bug is in an old tracker but still valid? IMHO this causes more work to keep the references up-to-date.

okurz:
https://github.com/os-autoinst/os-autoinst-needles-opensuse/pull/392 done for deleting all needles so subsequent tests should work

PR #391, opened 2018-07-04
[WIP] Fix needles for self-test

nicksinger:
Includes the same test as #390 and provides a proof that Travis can pass if everything is fixed.
If you merge #390 first, this one should be mergable too.


nicksinger:
Maybe a nice idea to fix them in parallel but git gives me a hard time carrying over the fixes from the other branch… so just closing it. We can still fix the stuff after the test got merged and enabled.

PR #392, opened 2018-07-04
Delete all invalid needles (no json or no bugref in workaround)

okurz:
This is in preparation of
https://github.com/os-autoinst/os-autoinst-needles-opensuse/pull/390
introducing a self-check which should be used for all follow-up PRs to
ensure these incorrect needles are not introduced in the future.


okurz:
LGTM by @nicksinger

PR #394, opened 2018-07-04
Use py.test instead of plain python for test

okurz:
Example output:

```
tests/test.py F                                                                                                                                        [100%]

========================================================================== FAILURES ==========================================================================
_____________________________________________________________________ test_missing_parts _____________________________________________________________________

    def test_missing_parts():
        for needle in sorted(needles):
            jsonfile = os.path.join(needledir, needle + '.json')
            pngfile = os.path.join(needledir, needle + '.png')

            # Check for file existence
            assert os.path.isfile(jsonfile), "Needle '{}' is missing its JSON file!".format(needle)
>           assert os.path.isfile(pngfile), "Needle '{}' is missing its PNG file!".format(needle)
E           AssertionError: Needle 'foo' is missing its PNG file!
E           assert False
E            +  where False = <function isfile at 0x7f297d413950>('/home/okurz/local/os-autoinst/os-autoinst-distri-opensuse/products/opensuse/needles/tests/../foo.png')
E            +    where <function isfile at 0x7f297d413950> = <module 'posixpath' from '/usr/lib64/python3.6/posixpath.py'>.isfile
E            +      where <module 'posixpath' from '/usr/lib64/python3.6/posixpath.py'> = os.path

tests/test.py:26: AssertionError
================================================================== 1 failed in 0.18 seconds ==================================================================
```


nicksinger:
Is it intentional that we still call the test by executing `python3 tests/test.py` with this PR? Or do you actually want to call `cd tests && py.test`? If the later applies, you also need to adjust [the script-line of .travis.yml](https://github.com/os-autoinst/os-autoinst-needles-opensuse/blob/master/.travis.yml#L3)

okurz:
true, true

asdil12:
Anyway this is the wrong repo now as the testsuite is centrally maintained at
https://github.com/os-autoinst/os-autoinst-needles-tests

PR #396, opened 2018-07-11
Needles for dolphin file operations test

Vogtinator:


asmorodskyi:
word "stuff" does not bring any additional meaning so can be dropped from all needle names / file names

asmorodskyi:
ok , I get it it is name of folder which you creating

PR #411, opened 2018-08-09
Fix needles for multi_user_dm test

asdil12:


displaymanager-gdm-user-prompt-20180809.json
```diff
@@ -0,0 +1,30 @@
...
+    {
+      "type": "match",
+      "xpos": 383,
+      "height": 77,
+      "width": 247,
+      "ypos": 656
+    }
+  ],
+  "tags": [
+    "displaymanager",
```
asmorodskyi: what about 
"ENV-DESKTOP-gnome",
"ENV-DISTRI-opensuse" ? also maybe something to filter out from versions which would not be used ?

asdil12:
updated patch

PR #412, opened 2018-08-10
Add SLE15 style system roles for openSUSE

jknphy:
https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/5551


jknphy:
Good idea! what about this error in the new CI? I rebased but I am not sure if I have to do something else. Could be produced for needles created on the fly in openQA  and not via CI?

jknphy:
Fixed, removed wrong workaround needle in a second commit.

PR #414, opened 2018-08-16
Make needle displaymanager-gdm-nomouseover-20180516 more specific to

ggardet:
avoid double matching and fix shutdown test for
opensuse-Tumbleweed-DVD-aarch64-Build20180813-create_hdd_gnome@aarch64:
  https://openqa.opensuse.org/tests/725135#step/shutdown/3


ggardet:
Tested locally with os-autoinst, shutdown test passes for AArch64.

rwx788:
May I ask you how you got the values for the needle? I usually create a new one, so it's more transparent if some job starts to fail.

ggardet:
@rwx788 The needle got renamed to spot it easily in openQA if things start to fail somewhere.
For the value, I simply increased the height area to include the "Not listed?" tag. You can use an image editor such as gimp to check the wanted value.

PR #416, opened 2018-08-17
Fix gnucash doesn't close when sending ctrl+q

SergioAtSUSE:
See code changes: https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/5595


SergioAtSUSE:
The needle was created here: https://github.com/os-autoinst/os-autoinst-needles-opensuse/commit/e825e3aca36c81e60abb42f066e25e4ac3ee7d77

PR #429, opened 2018-09-03
Make needle displaymanager-gdm-nomouseover-20171229 more specific

ggardet:
to avoid double matching and fix ensure_unlocked_desktop function used in
updates_packagekit_gpk for Leap 15.1:
https://openqa.opensuse.org/tests/745794#step/updates_packagekit_gpk/6

This is similar to what has been done for Tumbleweed in https://github.com/os-autoinst/os-autoinst-needles-opensuse/pull/414


ggardet:
Tested locally with os-autoinst.

PR #433, opened 2018-09-06
Additional checks for yast2 bootloader tabs

mloviska:
https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/5726


jknphy:
Really minor thing, but did you notice some "loading effect" in the third tab? if not, I guess with only one area picking the tab selected would have been enough.

PR #435, opened 2018-09-11
Replace grub2 boot loader needle

mloviska:
VRs:

* [opensuse-15.1-DVD-x86_64-Build301.2-yast2_gui@64bit](http://dhcp128.suse.cz/tests/6070#step/yast2_bootloader/5)
* [opensuse-Tumbleweed-DVD-x86_64-Build20180910-yast2_gui@64bit](http://dhcp128.suse.cz/tests/6071#step/yast2_bootloader/5)


okurz:
Can you state what is the motivation for this change? Does this fix failures? Is there a corresponding ticket?

mloviska:
It should be done as part of https://progress.opensuse.org/issues/39215. Before we were looking for arrow only https://openqa.opensuse.org/tests/719515#step/yast2_bootloader/5, which can be a possible point of failure in future, if there will be added another list option.

PR #437, opened 2018-09-12
Handle network manager default in openSUSE

jknphy:
https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/5752


okurz:
looks ok, however keep the comment in the ticket and test PR in mind: TW should also have network manager as default

jknphy:
We forgot to merge the needles, if someone can also merge them, thanks ;)

PR #438, opened 2018-09-12
Add workaround for missed wallpaper on Wayland

OleksandrOrlov:
Due to bsc#1105798, the wallpaper is missed on Wayland. The commit adds
the appropriate needles as a workaround until the issue is fixed.

Related ticket: [poo#40136](https://progress.opensuse.org/issues/40136)
Verification Run: http://oorlov-vm.qa.suse.de/tests/112#step/start_wayland_plasma5/21

**For reviewers:** The bsc#1105798 is marked as RESOLVED FIXED, but as per fvogt the fix is in :Update, but openQA doesn't use that.

So, I've added the workaround as the failing module blocks my investigations on other modules after it.


generic-desktop-kde-bsc1105798-20180824.json
```diff
@@ -0,0 +1,31 @@
+{
+  "properties": [
+    "workaround",
```
rwx788: I guess single property is enough =) Please remove and we can merge

SergioAtSUSE: But, this way you are sure ;)

okurz: but seriously, how did you manage to get 4 times the tag into here? I guess you did not do this manually but over the webUI, I guess this is an openQA bug then?

OleksandrOrlov: Haha, strange. Of course, I did that via webUI.

PR #442, opened 2018-10-03
Adjust YaST2 Samba after logon drive

jknphy:
https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/5874


rwx788:
Travis fails with `Needle 'yast2_bootloader-yast2-bootloader_grub2_for_EFI-20181002' includes a workaround tag but has no bug-ID in filename!` Which is not modified in this PR

PR #444, opened 2018-10-08
Fix problem with closing vlc by assert_and_click

Zaoliang:
even send_key_until_needlematch 'generic-desktop', "ctrl-q", 10, 10;
won't work realiable, we try now with assert_and_click and create required
needle 'close_app'.

delete vlc-url_check-20180614 because it breaks test and the older needle
vlc-url_check-20180515 works fine:
http://e13.suse.de/tests/9354#step/vlc/23

see ticket https://progress.opensuse.org/issues/37713 for details
verification run: http://e13.suse.de/tests/9354#next_previous


asdil12:
http://e13.suse.de/tests/9406#step/vlc/26

PR #448, opened 2018-10-11
Replace generic name close_app by close_vlc

Zaoliang:
delelte close_app needle and add close_vlc
add a new vlc-url_check which works reliable
verification run:
http://e13.suse.de/tests/9463#step/vlc/23
see ticket: https://progress.opensuse.org/issues/37713


okurz:
thanks.

PR #453, opened 2018-10-18
Delete wrong needles vlc-done-playing and create a correct one for vlc

Zaoliang:
we have problem with video playing, but existing needle could not detect
and check this issue. So create a new one at place where video is playing
for the case that video cannot be playing and vlc shows video information
only, we add record_soft_failure for this and it requires a new needle
vlc-video-info-only

see ticket https://progress.opensuse.org/issues/37713 for more details

verificatiion run:
http://e13.suse.de/tests/9652#step/vlc/25
http://e13.suse.de/tests/9640#step/vlc/25


okurz:
Please be careful to not miss any job depending on e.g. desktop environments, etc. . Or point out this issue to usual stakeholders reviewing tests on o3 that could run into this issue.

okurz:
I strongly recommend to rethink the choice of the needle as visible in http://e13.suse.de/tests/9652#step/vlc/25 . I can not imagine how this could be stable and also it does not show a VLC player being "done". Why not wait for the pause button to turn into the play button again?

PR #456, opened 2018-10-19
remove ref bsc1069458 in needle name

michelmno:
to avoid related test reported as soft-failed, eg:
https://openqa.opensuse.org/tests/777794#step/yast2_proxy/17


okurz:
https://bugzilla.suse.com/show_bug.cgi?id=1069458 is not VERIFIED FIXED so I assume we need to verify the job first before we remove the soft-fail, or not?

michelmno:
the bug is already in resolved fixed.  (not verified fixed)
but anyway my proposal is not to remove the needle, but  to rename it as will continue to be used as per source (yast2_proxy-port_opened tag) https://openqa.opensuse.org/tests/777794/modules/yast2_proxy/steps/1/src  and the soft_failed  is explicitely reported if yast2_proxy_squid_cannot-open-interface tag is found.

So there is no need to have workaround property for this needle.

okurz:
Yes, that part is OK. Could you please set the bug to verified though with at best a link to openqa showing the fixed state?

okurz:
*bump*

michelmno:
no more occurence since 2 month, so assume useless needle, so no need to rename it.
close the PR.

PR #458, opened 2018-10-25
Add needles for RAID test on TW

rwx788:
Some needles didn't match on tumbleweed due to different icons.

See [poo#42128](https://progress.opensuse.org/issues/42128).


okurz:
Of course I wonder why we need new needles but also in these new needles I do not see any "match level" other than default. Maybe it would help to define a match level of say 80% to also match whatever we have for other products, in this case most likely Leap 15.1?

rwx788:
@okurz it doesn't guarantee that newly created needles will be that smart. I had this situation before as well. But here it's clearly different icons. As it's staging, it can be the reason for this behavior. I cannot find ENV tags to enable them for staging only.

PR #465, opened 2018-11-02
Add new YaST2 UI firewalld

jknphy:
https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/6113


jknphy:
I have to admit that I didn't expect so many needles, the number of needles could have been less (perhaps) using text mode, but nevertheless if we want stable tests we need to assert actions after sending keys and for example in this case the GUI doesn't have 'easy ways' for navigation, for example from the tree view to the central panel, etc. Besides that needles for Leap and TW are different which doesn't help neither.

rwx788:
@jknphy glad that you share this concern, and that's the price for a proper testing. That's actually the point I've discussed with @OleksandrOrlov that we don't want to get another test module where we will spend time on maintenance, as we plan to start using UI framework.

PR #477, opened 2018-11-20
Add workaround in root-console-rpc-msg-boo1116598-20181119.json

michelmno:
I forgot to set it in previous creation as reported in
https://openqa.opensuse.org/tests/800841#step/shutdown_ltp/13


okurz:
thx

PR #485, opened 2018-12-11
Add kontact fatal error for workaround: bsc#1105207

Zaoliang:
Add workaround for kontact fatal error: bsc#1105207

we have issue with fatal error reported in bsc#1105207
add now workaround for this issue:
https://openqa.opensuse.org/tests/812844#step/kontact/14
see https://progress.opensuse.org/issues/35589
verification runs(all successful):
http://f40.suse.de/tests/89#step/kontact/
needle PR:
os-autoinst/os-autoinst-needles-opensuse#485


okurz:
Should be boo#1105207 instead of bsc#110520

okurz:
I don't see the updated commit in https://github.com/os-autoinst/os-autoinst-needles-opensuse/pull/485/commits and also no updated PR description

okurz:
see my comment above

Zaoliang:
again, this is strange. I can see the updates in

https://github.com/Zaoliang/os-autoinst-distri-opensuse/commit/29fb5ec62e27763e69914f1590374497cb558bc5

okurz:
@Zaoliang the PR description is not automatically updated with the commit message

okurz:
Ok, I updated the subject line now myself

PR #494, opened 2019-01-07
Change match level of gnome-screenlock-password-Tumbleweed

alapha23:
Raise match level of this needle. 

On https://openqa.opensuse.org/tests/824730#step/shutdown/3 , user name was wrong but it was 98% similar.


okurz:
hm, ok. let's take it. But it might be a bit confusing if the *same needle* matched in before and it would not match anymore. We could make that more explicit by renaming the needle file with a new name, e.g. update the timestamp in the filename. Hint for next time ;)

PR #497, opened 2019-01-21
Change default match level for vlc-url-check

Zaoliang:
we have sometimes problem with matching vlc-url-check. Change it now
to 86% to make the needle robust.
see https://progress.opensuse.org/issues/39926#note-9
verification test runs:
http://f40.suse.de/tests/228#next_previous


okurz:
I don't see the need in creating yet another needle without cleaning up old ones, especially as https://openqa.opensuse.org/tests/overview?distri=kubic&distri=opensuse&version=Tumbleweed&build=20190115&groupid=1&failed_modules=vlc currently does not show any VLC related failures. Also, I *think* it would be beneficial to mask any blinking cursor with an exclude area.

Zaoliang:
fine to me, then close it

PR #501, opened 2019-01-23
Delete workaround property from needles without bugref

okurz:


nilxam:
https://github.com/os-autoinst/os-autoinst-needles-opensuse/pull/502

PR #504, opened 2019-01-26
Revert "Use available screenshot to assert an unfocused textbox"

okurz:
Reverts os-autoinst/os-autoinst-needles-opensuse#499

The needle is neither specific to Leap 15.0 nor is it restricted to match only an "unfocussed textbox", see https://openqa.opensuse.org/tests/839815#step/start_wayland_plasma5/32


okurz:
@rwx788 @SergioAtSUSE

rwx788:
I've restarted that job.

PR #511, opened 2019-02-08
Add tianocore needles for aarch64

ggardet:


okurz:
created two new needles for the according SLE test https://openqa.suse.de/tests/2446124

PR #517, opened 2019-02-25
Include install button to the matches

baierjan:
Fix for https://progress.opensuse.org/issues/48377

Verification run: http://panigale.suse.cz/tests/1462#step/updates_packagekit_gpk/19


okurz:
That was fast :) Thank you. The self-checks currently fail because of "generic-desktop-kde-tw-kde404380-20190215" which is not referencing a bug from our bugzilla instance. That is certainly unrelated :)

PR #554, opened 2019-05-06
Renew and add new needles for tests/console/gpg.pm

wnereiz:
Ticket: [poo#50606](https://progress.opensuse.org/issues/50606)

Test case PR in **os-autoinst-distri-opensuse**: [Refine GnuPG case with batch mode for stability](https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/7345)


gpg-enter-passphrase-reenter-20190506.json
```diff
@@ -0,0 +1,17 @@
...
+      "width": 253,
+      "height": 18,
+      "type": "match",
+      "match": 75,
+      "margin": 150
+    }
+  ],
+  "properties": [],
+  "tags": [
+    "gpg-enter-passphrase-reenter"
```
asmorodskyi: I get your point you want to follow same pattern , but for me it sounds too weird . maybe we fallback just to `gpg-reenter-passphrase` ?

asmorodskyi: or you can use as main pattern `gpg-passphrase` : 

- `gpg-passphrase-enter`
- `gpg-passphrase-reenter`
- `gpg-passphraseunlock`

gpg-insecure-passphrase-20170614.json
```diff
@@ -10,6 +10,7 @@
     }
   ],
   "tags": [
-    "gpg-incorrect-passphrase"
+    "gpg-incorrect-passphrase",
```
asmorodskyi: looking into screenshot , agree with that it is about insecure but I didn't get why you did addition not replacement ? incorrect passphrase is different test case in my understanding .

gpg-enter-passphrase-reenter-20190506.json
wnereiz: I followed your second suggestion, I have also change the test case.

gpg-insecure-passphrase-20170614.json
wnereiz: That is because if this PR merged before test case PR and there is case running during this period, it will failed to find the tag. But any way, I have removed it, please merge too PRs together.

PR #561, opened 2019-06-04
Add needles for installation on krypton

rwx788:
See [poo#52445](https://progress.opensuse.org/issues/52445).


rwx788:
@okurz first of all, please, take a look on the change before requesting changes. The only thing where we might have duplicates are the issues with typing. And I had really weird ones, having ">" instead of dot. Other needles are unique to #560.

okurz:
ok, let's merge your PR and I will rebase mine and see what is still necessary

PR #562, opened 2019-06-18
Cleanup dolphin_new_folder

StefanBruens:
dolphin_new_folder-20180915 matches at 80%, and lets cases pass
where the "stuff" was not entered correctly

dolphin_new_folder-20180712 is the same as 20180915, but without
the mask for the blinking cursor, thus redundant

dolphin_new_folder-20180710 is the same as 20180713, but without
the mask for the parent folder (which changes with the login name).


okurz:
@jknphy The checks fail because of another needle that has the workaround property but no bugref in the filename. I will fix that in another PR.

PR #563, opened 2019-06-19
Fix 'includes a workaround tag but has no bug-ID in filename' error in self-test?

okurz:


okurz:
@jknphy

jknphy:
needle await_install-package-broken-yast2-logs-20190617 deleted. Sorry for the mess, I should have deleted it, it was just a test. Retry please.

okurz:
ok so we do not need this PR anymore

PR #566, opened 2019-07-02
Delete needles which have problem with click coordinates

Zaoliang:
see https://openqa.opensuse.org/tests/973050/modules/gnucash/steps/32
related ticket: https://progress.opensuse.org/issues/50492


Zaoliang:
see my PR: https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/7662/files
tps://openqa.opensuse.org/tests/973050#step/gnucash/35 click of cancel doesn't work.

okurz:
your link is broken

okurz:
And also this can not be the problem. https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/7662/files#diff-a9d3fcafd40bc069165aaf96e875d7f5R27 is simply a call to `assert_screen` on gnucash-assistant-close, not `assert_and_click`, so what would you expect?

okurz:
https://openqa.opensuse.org/tests/971829#step/gnucash/25 is the last successful application of one of the needles you want to delete so I doubt this is a good idea.

Zaoliang:
please see http://f40.suse.de/tests/4166#step/gnucash/37:

display welcome dialog again, this coming up because the previous needle with correct coordinate for click works. old needle matches, yes, but it has problem with "click".
http://f40.suse.de/tests/4166#step/gnucash/35 shows the new needle which works for "click".

PR #599, opened 2019-10-11
New needles for PXE reboot after installation.

klwagner:
For modules installation/reboot_after_installation.pm and boot/boot_from_pxe.pm
https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/8538


okurz:
https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/8538 mentions "As of now this works only on SLES-15 SP1" so why should we need needles for SLE in os-autoinst-needles-opensuse

klwagner:
Interest in the referred test for openSUSE appears to be low at the time. Closing this PR.

PR #602, opened 2019-10-24
Check for Missing e2fsprogs bsc#1140040

Zaoliang:
see https://progress.opensuse.org/issues/56423
verification run:
http://f40.suse.de/tests/5254#step/kdump_and_crash/26


yast2-kdump-unexpected-issue-20191024.json
```diff
@@ -0,0 +1,16 @@
...
+    {
+      "xpos": 211,
+      "ypos": 334,
+      "width": 395,
+      "height": 98,
+      "type": "match",
+      "match": 86
+    }
+  ],
+  "properties": [],
```
SergioAtSUSE: The green frame of the screenshot is confusing. This needle should be a workaround needle.
Guideline: https://progress.opensuse.org/projects/openqatests/wiki#applying-quotworkaroundquot-needles

Zaoliang: updated now

SergioAtSUSE:
Something strange happened. I have noticed that the old file (with old filename and no woraround tag is also included in the commit) could you check and remove it?

Zaoliang:
yes, this is strange. We need to step by step in future, delete and create a new needle. just rename and edit this is not good ;)

SergioAtSUSE:
Well, when you rename a file, `git status` does not show it as renamed. It shows a removed file and an added file. So, `git rm $old_file_name` and `git add $new_file_name` is needed. Then, in staging, `git status` will show them as renamed if the content is still higher than 80% similarity, IIRC, the actual minimum percent may be different.

PR #612, opened 2019-11-19
Create new needles for virtman-view

Zaoliang:
we have problem with the virtman test module for long time.
send_key doesn't work realible and even trying with VNC_TYPING_LIMIT
could not help us to get this test module passed.
Use assert_and_click is working in this case then.

see https://progress.opensuse.org/issues/54056
verification test:
http://f40.suse.de/tests/5320#step/virtman_view/13


rwx788:
CI issue will be fixed in https://github.com/os-autoinst/os-autoinst-needles-opensuse/pull/613

rwx788:
@Zaoliang please re-trigger travis (e.g. by force push), irrelevant failure should be gone now.

Zaoliang:
@SergioAtSUSE no I don't want to use send_key for this module because the poor performance of virt-manager will have problem with send_key soon or later, thanks!

PR #613, opened 2019-11-19
Fix naming of the needle from commit 91ff3af70

rwx788:
Needle names without timestamp are not allowed and break CI.


rwx788:
@michelmno JFYI.

PR #619, opened 2019-11-28
Replace linux-login by tty1-selected for firstrun.pm

Zaoliang:
like tag text-login, replace linux-login by tty$Nr-selected.
see https://progress.opensuse.org/issues/34471
verification test:
http://f40.suse.de/tests/6772#step/console_reboot/4
http://f40.suse.de/tests/5579#step/firstrun/7


foursixnine:
How does this behave in other cases?, where tty6-selected (or root console, according to the code in the text module)? because this one, seems to expects the OK messages to be shown there, always... so seems to be something very specific to JEOS, also.. tty6 != tty1!

Zaoliang:
just check this tty$Nr:

my $tty = get_root_console_tty;
assert_screen ["tty$tty-selected", 'reached-power-off'], 1000;

Zaoliang:
http://f40.suse.de/tests/5579#step/firstrun/7
this is also updated in commit message.

Zaoliang:
I opened now for tty$Nr detection issue:
https://progress.opensuse.org/issues/60431

SergioAtSUSE:
> There seems to be changes missing on the test side of things according to this http://f40.suse.de/tests/5579/modules/firstrun/steps/1/src.
> 
> On another hand I wonder why jeos's console is not the right one (`my $tty = get_root_console_tty;` should return the right & expected one) from the beginning. @Zaoliang Could you figure that out as a follow up? /cc @SergioAtSUSE

So, as already two person, including me, get confused by seeing only a PR with needles that should be a dependency of code change PR, @Zaoliang, please open also always the PR with the code changes, and put a link the the needle's PR in the description.

SergioAtSUSE:
> On another hand I wonder why jeos's console is not the right one (`my $tty = get_root_console_tty;` should return the right & expected one) from the beginning. @Zaoliang Could you figure that out as a follow up? /cc @SergioAtSUSE

If you mean, "why is get_root_console_tty returning two different consoles?":
https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/utils.pm#L1055

If you mean, "why is tty1 being shown?", this is why I complained that we shouldn't expect root-console (in this case tty6) here, because we are at that point only checking if the system booted. The select_console statement happens later and then should be check for tty6 (or tty2, seeing get_root_console_tty() code)

Zaoliang:
anyway, this might be not a problem with detection of tty at this stage, but as reference which can be expected:
http://f40.suse.de/tests/5580#step/firstrun/10
so we had already tty1-selected for sles jeos.

PR #623, opened 2019-12-08
Add needles for new gnuhealth 3.6 in Tumbleweed

okurz:
Related test change:
https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/9101

Related progress issue: https://progress.opensuse.org/issues/60740


asmorodskyi:
from what I understand from test code which will use this needles , they are relevant only for Leap 15.2 , right ? In case my assumption is right maybe we need to put version tags to this needles ?

okurz:
no, the subject also states Tumbleweed, no version. And we should never add a version tag except you mean something like ENV-SP2ORLATER-1

PR #625, opened 2019-12-13
Add needles for GraphicsMagick test

ilausuch:
Needles used by os-autoinst/os-autoinst-distri-opensuse#9163


graphics_magick-cleared-console-20191213.json
```diff
@@ -0,0 +1,15 @@
...
+      "width": 809,
+      "height": 524,
+      "type": "match"
+    }
+  ],
+  "properties": [],
+  "tags": [
+    "cleared-console"
+  ]
+}
```
okurz: I don't see where this needle should be needed. It has a generic tag "cleared-console" but with a filename stating it would be explicit for "graphics_magick". I know, the openQA needle editor suggests this filename but we should not use that for generic tags. Maybe you are missing an update to needles from the main repo?

graphics_magick-Open_an_images_histogram-20191213.json
```diff
@@ -0,0 +1,15 @@
...
+      "xpos": 1,
+      "ypos": 68,
+      "width": 196,
+      "height": 193,
+      "type": "match"
+    }
+  ],
+  "properties": [],
+  "tags": [
+    "Open_an_image_histogram"
```
okurz: please use lowercase tags only which is a more common approach. Applies to all needles.

graphics_magick-cleared-console-20191213.json
ilausuch: Maybe this is the cause. While I was creating the needles this suggested me this needle. Checking...

graphics_magick-Open_an_images_histogram-20191213.json
ilausuch: ok, changing

graphics_magick-cleared-console-20191213.json
ilausuch: Done

graphics_magick-Open_an_images_histogram-20191213.json
ilausuch: Done

ilausuch: Done

PR #631, opened 2020-01-24
Fix opensuse_welcome-kde-20200122-notification-area-excluded naming (…

DimStar77:
…date at the end)


DimStar77:
Actually, discarding - the needle was wrong from the get-go. It was masking a bug; I thus removed it from the needles list using OBS needle UI

PR #633, opened 2020-02-06
Prevent false matches of 'screenlock-xscreensaver'

okurz:
Related progress issue: https://progress.opensuse.org/issues/48899


ggardet:
It breaks a number of tests, mainly for GNOME tests on aarch64 (Tumbleweed and Leap 15.2).
See: https://progress.opensuse.org/issues/48899#note-6

PR #640, opened 2020-02-13
Rename tags for raid partitioning needles

sofiasyria:
Renaming tags for raid partitioning needles. The new tags are according to format "partitioning_%s-format-selected" in order to match the changes of pull request #9237 : https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/9237/files#diff-3a91353c18f2ab16cdd0abc6b2c927b5

### DIFF
--- a/partitioning_raid-fat_format-selected-20190125.json
+++ b/partitioning_raid-fat_format-selected-20190125.json
@@ -10,6 +10,6 @@
   ],
   "properties": [],
   "tags": [
-    "partitioning_raid-fat_format-selected"
+    "partitioning_fat-format-selected"
   ]
 }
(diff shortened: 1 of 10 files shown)


okurz:
Please follow https://commit.style/ or http://chris.beams.io/posts/git-commit/ as a helpful guide how to write good commit messages.

The subject should be written in imperative mood and no dot at end of line.

partitioning_raid-fat_format-selected-20190125.json
```diff
@@ -10,6 +10,6 @@
   ],
   "properties": [],
   "tags": [
-    "partitioning_raid-fat_format-selected"
+    "partitioning_fat-format-selected"
```
jknphy: I think we should be consistent with name of the files and needle tag, wdyt? I see you delete the raid part so maybe worth to do the same in the name of the file, .png and .json, but we don't care too much about names, so up to you, it is not like you have to do it :)

PR #642, opened 2020-02-19
Add needle for :// in chromium urlbar selected

asdil12:
Ticket: https://progress.opensuse.org/issues/63334


chromium-urlbar-colon-alash-20200219.json
```diff
@@ -0,0 +1,15 @@
...
+      "xpos": 207,
+      "ypos": 70,
+      "width": 12,
+      "height": 19,
+      "type": "match"
+    }
+  ],
+  "properties": [],
+  "tags": [
+    "chromium-urlbar-colon-alash"
```
SergioAtSUSE: This should be actually called **chromium-hightlighted-urlbar**

asdil12: updated: http://artemis.suse.de/tests/2266#step/chromium/19

PR #646, opened 2020-02-21
Delete 'screenlock-xscreensaver' yielding lots of false-positives

okurz:
The bad needle "screenlock-xscreensaver-20180821" causes even more problems.
E.g. https://bugzilla.opensuse.org/show_bug.cgi?id=1164311 was reported with
something about an error message which has nothing to do with the test failure
which is very related to the bad needle causing false-positive matches. Where
necessary a much stricter needle should be provided instead, e.g. with a match
level of 100.

Related progress issue: https://progress.opensuse.org/issues/48899


Zaoliang:
this needle has second tag 'blackscreen'.
tests/installation/await_install.pm uses tag 'blackscreen' and there is no needle which contains it atm.
I'm afraid the delete of this needle could effect await_install.pm

ggardet:
@okurz A lot of failure due to this deletion happens again. For example on Leap 15.2 ARM:
https://openqa.opensuse.org/tests/1183424#step/consoletest_finish/1
https://openqa.opensuse.org/tests/1183433#step/x11/2
https://openqa.opensuse.org/tests/1183763#step/consoletest_finish/12
https://openqa.opensuse.org/tests/1183762#step/consoletest_finish/12

I think we should keep this needle, but check for the `blackscreen` tag when required?

okurz:
@ggardet @Zaoliang No way! The needle *is* causing false-positives because it was simply not strict enough matching on the console prompt which is *clearly* not a black screen

okurz:
@Zaoliang yes, this is accounted for. The needle deleted here was matching on a screen that was *not* a black screen as well causing e.g. the module you mentioned "await_install" to be stuck in an endless loop causing the jobs to timeout when they should fail. Deleting this needle doesn't magically fix tests but makes test failures more obvious. If you are interested in more details please read https://progress.opensuse.org/issues/48899 and https://progress.opensuse.org/issues/63346 carefully.

ggardet:
@okurz but removing it _causes_ tests to fails. 

I think we have 2 options here:
a) Revert this PR and add a `match_has_tag('blackscreen')` check where we know it can causes problems. (Maybe adjust match level to 100% and/or add additional needles to avoid false positive)
b) Re-introduce it as `blackscreen` only (not screenlock) with a match of 100% and check explicitly for this blackscreen on required tests, including in unlock desktop. But we will probably go back to the previous behavior and need to implement a) anyway.

So, I really think a) is the best way to go. I will try to work something out.

okurz:
I don't see reverting as the option but you can try to disable the black screen completely as it is so ambiguous. If all else fails you can try with a "blackscreen" needle but please set match level to 100% and this might not even be strict enough for a full screen needle.

PR #647, opened 2020-03-02
Add openSUSE license needles for JeOS

mloviska:
https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/9670


mloviska:
I will re-needle in o3.

PR #651, opened 2020-03-10
Add needles for rstudio tests for Leap 15.2

dcermak:
This is a full re-creation of the needles for the tests in the `extra_tests_rstudio` to be compatible with Leap 15.2. The main differences are:
1. slightly wider fonts
2. Firefox has a different title bar


okurz:
I fear this needle set will be hard to maintain. Also it seems not a single needle uses a match level different than default. Can you try to lower the match level to share needles between Leap and Tumbleweed and make needles more resilient?

dcermak:
Hm, it could work for some of them, but in the `rstudio_server-*` needles also have the issue that Firefox looks quite different on Leap vs on Tumbleweed.

Do you maybe have a rough idea how low the match level has to be set? I've never experimented with that and if you got some experience with that, it could save me quite some time.

tjyrinki:
This is an old MR that somehow has not been merged or discussed further. Please resubmit if still needed.

PR #657, opened 2020-04-28
add systemd-testsuite-binary-tests-summary

eliroca:
One is for systemd v244 and the other for v245. Do we really need this needle? Seems to be changing quite often.


eliroca:
Failed run: https://openqa.opensuse.org/tests/1238964#step/binary_tests/23
Successful run: http://emerson.suse.de/tests/2147#step/binary_tests/35

foursixnine:
@eliroca I'm actually wondering why we're matching needles for a console test, in any case, I would suggest not matching the numbers of the tests, but only the headers of the table (two match areas perhaps?) or even better, save_screenshot so that the test run results are recorded in a screen and that's it.

eliroca:
I'd take this needle check out, as the result is checked with `validate_script_output`.

foursixnine:
Just save a screenshot :) for the lazy

eliroca:
Then let's close this pull for now

PR #673, opened 2020-07-16
Delete plasma-mpris-pause-20191114, use a more specific one

Vogtinator:
This needle doesn't include the artwork, which is a sign that the applet
is fully loaded and can be interacted with. As a result, clicks got lost
during shader compilation (boo#1172763)

Verification runs:
https://openqa.opensuse.org/tests/1334385
https://openqa.opensuse.org/tests/1334386


Vogtinator:
cc @a-kpappas for the needle name fixes to make travis happy again.

PR #684, opened 2020-08-20
Add some needles for adjustment of seahorse

Zaoliang:
we need some new needles for adjustment and record_soft_failure.
create a workaround needle because of bsc#1175513.
see https://progress.opensuse.org/issues/66149
verification:
http://10.162.23.47/tests/8005#step/seahorse


SergioAtSUSE:
Please, remove seahorse-launched-20200427
and create a new needle with workaround property ([bsc#1175513](https://bugzilla.opensuse.org/show_bug.cgi?id=1175513)) and tag seahorse.

See https://progress.opensuse.org/projects/openqatests/wiki#applying-workaround-needles

PR #691, opened 2020-09-10
Add two needles for ppc64le microos tests

michelmno:


michelmno:
will do hit manually via o3 ui, so close this pr

PR #695, opened 2020-10-21
Add WSL2 update needles

mloviska:
https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/11253


b10n1k (changes requested):
Could you add a new line in each .json?

mloviska:
> Could you add a new line in each .json?

Never done that, does it help anything? Shouldn't it be fixed in openQA?

PR #703, opened 2020-11-25
Adapt disk activation for Tumbleweed on non-standard 150 dasd path

SergioAtSUSE:
- Verification run: http://openqa.slindomansilla-vm.qa.suse.de/tests/2880#step/disk_activation/8


SergioAtSUSE:
Self-merging since they are needles, I could have created them via the webui without review and verification run looks fine.

PR #710, opened 2021-02-01
Add Xfce mainmenu opened with the super key

Vogtinator:
As opposed to the previous popup menu.


Vogtinator:
Pushed manually to avoid Travis

PR #711, opened 2021-02-02
Add needles for Thunderbird on openSUSE Tumbleweed

martinsmarcelo:
This needles are necessary to run on the  last Thunderbird versions, and with multimachine scenarios .


okurz:
As a SUSE employee please take care to use the correct way to write *openSUSE*

PR #720, opened 2021-04-26
Reduce match percentage for inst-rootpassword as cursor causes failure

sofiasyria:
The cursor, when percent in the matching area is causing test failure. Reducing the match percentage to 85% should fix that.

https://openqa.opensuse.org/tests/1714886#step/yast2_firstboot/12


okurz (changes requested):
Please mask the mouse cursor with an exclude area instead

sofiasyria:
> Please mask the mouse cursor with an exclude area instead
ok

PR #723, opened 2021-05-10
Adds needle for ubuntu login

b10n1k:
https://progress.opensuse.org/issues/88822


Vogtinator:
IMO this should have a `ENV-UBUNTU-1` or something like that to avoid that it is used in other tests.

Vogtinator (approved):

b10n1k:
> IMO this should have a `ENV-UBUNTU-1` or something like that to avoid that it is used in other tests.

See also a4d1ac86471294c8840d698f31ea9664fff4bcbb from https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/12507/commits

PR #725, opened 2021-05-17
Delete needles impacted by boo#1186093

Vogtinator:
A bug in the font generation script caused missing line drawing glyphs.


Vogtinator:
As this required some new needles anyway, I just replaced the affected ones using the WebUI and deleted the inst-welcome-textmode one instead..

PR #733, opened 2021-06-29
Replace boo#1139927 reference with updated boo#1137203

okurz:
https://bugzilla.suse.com/show_bug.cgi?id=1139927 is "VERIFIED
DUPLICATE" of https://bugzilla.suse.com/show_bug.cgi?id=1137203 . Also
it seems the problem now only appears anymore on aarch64.


okurz:
@asdil12 you had been involved with the investigation in https://bugzilla.suse.com/show_bug.cgi?id=1137203 so please review.

@ggardet I have seen https://openqa.opensuse.org/tests/1755968#step/tk/10 still making problems but https://bugzilla.suse.com/show_bug.cgi?id=1137203 is set to "resolved". I recommend to reopen with a reference to current tests. Otherwise likely we will have reminder comments after accepting this PR anyway.

okurz:
As it turns out the problem is actually not limited to aarch64 so I updated again to account for that.

asdil12:
It looks like on x86_64 the text is actually shown but an updated needle is missing:
https://openqa.opensuse.org/tests/1757267#step/tk/10

asdil12 (approved):

okurz:
> It looks like on x86_64 the text is actually shown but an updated needle is missing:
> https://openqa.opensuse.org/tests/1757267#step/tk/10

you are right. Did you create a new one?

asdil12:
> you are right. Did you create a new one?

yes

PR #749, opened 2021-09-17
Replace broken travis CI with github actions

asdil12:


okurz (changes requested):
you included a lot of unnecessary boiler-plate code in the GHA definitions. I suggest to only include the relevant parts and also remove all unnecessary comments.

Can you also specify in more details what you consider "broken" that you want to fix?

asdil12:
When I created that PR I actually thought that travis was completely broken. Afterwards I saw that the result eventually appeared, which still took quite a while.

asdil12:
> you included a lot of unnecessary boiler-plate code in the GHA definitions. I suggest to only include the relevant parts and also remove all unnecessary comments.
> 
> Can you also specify in more details what you consider "broken" that you want to fix?

updated PR

okurz (changes requested):
The commit message could have a bit more details, e.g. what issue is fixed, why this needs to change, to which versions of which product it applies, link to a bug or a fate entry, the choices you made, etc. Also see https://commit.style/ or http://chris.beams.io/posts/git-commit/ as a helpful guide how to write good commit messages.

asdil12:
This is how the test suite result will look like with the current yaml:
https://github.com/asdil12/os-autoinst-needles-opensuse/runs/3649491718?check_suite_focus=true

asdil12:
I updated the commit but there is no ticket.
I just realized that os-autoinst was complaining about duplicate tags within needles and this was not covered, when working on https://progress.opensuse.org/issues/97751.
Then I found out that there are other issues that the testsuite should have had covered but it slipped through.
If you look at the commits on the master branch, travis doesn't even seem to run on commits from the openQA webui - for whatever reason.

okurz (approved):
I wouldn't mind seeing all what you wrote in your latest comment in the git commit but you just put there "github actions because github" but nevermind :D

PR #769, opened 2022-06-20
New needles required for the module wsl/install_from_MSStore

pablo-herranz:


grisu48:
Is this really what we want? I think those belong into the SLES needle repository

PR #792, opened 2023-01-23
ci: Add a retry loop around package installation

kalikiana:
Also a job-based retry of 2 (the maximum) in case the job fails for other reasons.

This addresses common errors like the following:

- Cannot read input: bad stream or EOF.
- Media source [...] does not contain the desired medium

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -4,12 +4,18 @@ stages:
   - mirror
 
 mirror-from-github:
+    retry: 2
     stage: mirror
     rules:
           - if: $TOKEN
     before_script:
         - source /etc/os-release && zypper ar http://download.suse.de/ibs/SUSE:/CA/openSUSE_Leap_${VERSION}/SUSE:CA.repo
-        - zypper -n in git ca-certificates-suse
+        - |
+          for i in {1..3}; do
+              zypper -n in git ca-certificates-suse
+              zypper ref
+              sleep 1
+          done
     script:
         - git config user.email "you@example.com"
         - git config user.name "Your Name"
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -12,8 +12,7 @@ mirror-from-github:
         - source /etc/os-release && zypper ar http://download.suse.de/ibs/SUSE:/CA/openSUSE_Leap_${VERSION}/SUSE:CA.repo
         - |
           for i in {1..3}; do
-              zypper -n in git ca-certificates-suse
-              zypper ref
+              zypper -n in git ca-certificates-suse && zypper ref && break
               sleep 1
           done
     script:


.gitlab-ci.yml
```diff
@@ -4,12 +4,18 @@ stages:
...
     rules:
           - if: $TOKEN
     before_script:
         - source /etc/os-release && zypper ar http://download.suse.de/ibs/SUSE:/CA/openSUSE_Leap_${VERSION}/SUSE:CA.repo
-        - zypper -n in git ca-certificates-suse
+        - |
+          for i in {1..3}; do
+              zypper -n in git ca-certificates-suse
+              zypper ref
+              sleep 1
```
Martchus, line 17: It is bad to sleep 1 second *unconditionally*, even when everything works fine. This loop will also *always* make 3 attempts, even if none of the commands in it fail. Supposedly just a break in case the commands succeeded is missing:

```suggestion
              zypper -n in git ca-certificates-suse && zypper ref && break
              sleep 1
```

Or should `zypper ref` only be done when the first command failed? Like this:

```suggestion
              zypper -n in git ca-certificates-suse && break
              zypper ref
              sleep 1
```

Martchus (changes requested):

Martchus (approved):

kalikiana:
Unrelated error (I'm only touching GitLab CI here):
```
Run ./tests/test.py
Needle 'system_role-system-role-textmode-selected' missing or invalid timestamp!
Error: Process completed with exit code 1.
```

.gitlab-ci.yml
```diff
@@ -4,12 +4,17 @@ stages:
...
+    retry: 2
     stage: mirror
     rules:
           - if: $TOKEN
     before_script:
         - source /etc/os-release && zypper ar http://download.suse.de/ibs/SUSE:/CA/openSUSE_Leap_${VERSION}/SUSE:CA.repo
-        - zypper -n in git ca-certificates-suse
+        - |
+          for i in {1..3}; do
+              zypper -n in git ca-certificates-suse && zypper ref && break
```
perlpunk, line 15: So the refresh is only done when the install passes?

mimi1vx: yeah, and then if refresh passes -> loop is stooped. It pretty clever solution , only `--force` in refresh will be nicer

perlpunk: but we should only need to refresh if the install *fails*, right?

perlpunk: Are there any plans to fix this? Is there a ticket?

Martchus: I also wasn't sure about this. That's why I proposed it as alternative in my previous comment (when requesting changes).

PR #795, opened 2023-04-14
Delete all needles which were not used since 2022

Vogtinator:
List of needles collected and then deleted as follows:

```
psql openqa -t -c "SELECT filename FROM needles WHERE file_present IS TRUE GROUP BY filename HAVING MAX(last_matched_time) IS NULL OR EXTRACT(YEAR FROM MAX(last_matched_time)) < 2022" > needle_cleanup
xargs rm < needle_cleanup; sed 's/json$/png/' needle_cleanup | xargs rm
```


Vogtinator:
The cutoff date is open for discussion. I don't think we'll need needles which are older than than min(a year ago, last full Leap test run) though. And if we do a cleanup, might as well do it properly. They're still in the git history, after all.

Last matched < 2022: 6191
Last matched < 2021: 4339
Last matched < 2020: 2490
Last matched < 2019: 483
Last matched < 2018: 483 (probably just all "never matched" needles)

okurz:
I read "since 2022" as "only 2023 needles" but I read the SQL statement as any match last in 2021 or older, right? If yes, then +1

Vogtinator:
> I read "since 2022" as "only 2023 needles" but I read the SQL statement as any match last in 2021 or older, right? If yes, then +1

Yes, I could've written "not used after 2021" as well.

okurz (approved):

Vogtinator:
Let's just give this a try.

PR #812, opened 2024-05-13
pvm bootloader & vnc installation specific needle

sneh-3:
@foursixnine


foursixnine:
@sneh-3 I need to check wrt the SLES needles in opensuse, I'll get back to you sometime next week on this (As I think we can also share our sles needles so you don't have to redo the work)

PR #820, opened 2024-09-09
Update validate_product-root-console-leap160-20240909.json

lkocman:


nilxam:
I'll do cleanup on openqa directly

PR #825, opened 2025-01-23
Needles for VNC as non-root

Vogtinator:


Martchus (approved):
Not sure how this is going to be used, though.

Vogtinator:
> Not sure how this is going to be used, though.

https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/21031, but waited after lunch for VR completion

PR #837, opened 2025-05-08
Add needles for selecting the TPM2+PIN method

mtravitzky:
Add needles for selecting the TPM2+PIN method in jeos-firstboot. Additionally, needle file names have been fixed (missing dash before date).


encrypted-disk-password-prompt-tpm-20250507.json
```diff
@@ -0,0 +1,15 @@
...
+      "xpos": 1,
+      "ypos": 738,
+      "width": 438,
+      "height": 29,
+      "type": "match"
+    }
+  ],
+  "properties": [],
+  "tags": [
+    "encrypted-disk-password-prompt"
```
Vogtinator, line 13: Should probably be a different tag to show PIN vs. password entry

Vogtinator (changes requested):
Needs a separate tag

encrypted-disk-password-prompt-tpm-20250507.json
mtravitzky: Fixed.

Vogtinator (approved):

Martchus (approved):

mtravitzky:
> We normally avoid having merge commits on feature branches. Can you just rebase against master? It would perhaps also make sense to squash the commits.
> 
> And do you need help merging this (as this PR has been sitting here now for 3 weeks)?

Rebased and squashed. Should be okay to merge now.

PR #838, opened 2025-05-08
Delete all needles with no or last match before 2024

Vogtinator:
psql openqa -t -c "SELECT filename FROM needles WHERE file_present IS TRUE GROUP BY filename HAVING MAX(last_matched_time) IS NULL OR EXTRACT(YEAR FROM MAX(last_matched_time)) < 2024" > needle_cleanup_2025


Vogtinator:
Hm, there's a weird `+1` in the diffstat I need to fix

okurz (approved):

PR #842, opened 2025-06-10
Fix tag in needle to match grub-bls

foursixnine:


Vogtinator:
IMO the code must not rely on a needle match there. GRUB2 with and without BLS might look completely identical.

foursixnine:
> IMO the code must not rely on a needle match there. GRUB2 with and without BLS might look completely identical.

Yeah, perhaps this part will need some rework, I guess the most troublesome module is bootloader_uefi

PR #844, opened 2025-08-05
Add needles for x11/xosview.pm

janvhs:
This commit adds needles for the functionality test of xosview.  It was first introduced in efefa878ed3bb1bd84d7b052696c0e0b4b8fbcdf of https://github.com/os-autoinst/os-autoinst-distri-opensuse.


Martchus:
I hope the area is outside the figures that are likely different between test runs.

d3flex (changes requested):
there is a double dash in the names. can we fix it?

